PR #5, opened 2024-07-02
Bug fixes

kirilklein:
- Fixed config, previously resulting in an error at instantiation
- Adopted new naming batch_size->batchsize
- Consistent use of lowercase and uppercase column names (lowercase for unprocessed and uppercase for final columns): this also resulted in an error.


corebehrt/common/config.py
```diff
@@ -1,25 +1,26 @@
...
 import yaml
-
+import json
+from os.path import join
 
 class Config(dict):
     """Config class that allows for dot notation."""
     def __init__(self, dictionary=None):
-        super().__init__()
+        super(Config, self).__init__()
```
mikkelfo, line 9: This is not needed, we can just do `super().__init__()`

And we probably need to have `super().__init__(dict)` and we can leave the below code out (?)

corebehrt/functional/creators.py
```diff
@@ -9,26 +9,26 @@
                               get_time_difference, normalize_segments_df)
 
 
-def create_ages(concepts: pd.DataFrame, birthdates: dict) -> pd.DataFrame:
+def create_ages(concepts: pd.DataFrame, patients_info: pd.DataFrame) -> pd.DataFrame:
```
mikkelfo, line 12: I think it's more expressive to have the required inputs for the function in `functional`, rather than passing in `concepts` and `patients_info`

I think we should have `def create_ages(timestamp: pd.Series, birthdates: pd.Series)` and handle the other things outside the `functional` library (perhaps inside `classes`)

corebehrt/functional/creators.py
```diff
@@ -9,26 +9,26 @@
...
     """Creates the ABSPOS column"""
-    concepts['ABSPOS'] = get_abspos_from_origin_point(concepts['TIMESTAMP'], origin_point)
+    concepts['abspos'] = get_abspos_from_origin_point(concepts['TIMESTAMP'], origin_point)
     return concepts
 
 def create_segments(concepts: pd.DataFrame, segment_col='ADMISSION_ID') -> pd.DataFrame:
     """ Creates the SEGMENT column (the normalize segments_df can do this) """
-    concepts['SEGMENT'] = normalize_segments_df(concepts, segment_col)
-    return concepts
+    return normalize_segments_df(concepts, segment_col)
```
mikkelfo, line 25: This doesn't follow the format as the other create functions, where we create the new column of concepts inside the function - However, I think the other functions should follow this format

corebehrt/functional/utils.py
```diff
@@ -7,7 +7,7 @@ def normalize_segments(x: Union[pd.Series, pd.DataFrame, list, dict], segment_co
     if isinstance(x, pd.Series):
         return normalize_segments_series(x)
     elif isinstance(x, pd.DataFrame):
-        return normalize_segments_df(x)
+        return normalize_segments_df(x, segment_col=segment_col)
```
mikkelfo, line 10: Should force naming

corebehrt/functional/utils.py
```diff
@@ -16,7 +16,8 @@ def normalize_segments(x: Union[pd.Series, pd.DataFrame, list, dict], segment_co
         raise TypeError('Invalid type for x, only pd.DataFrame, list, and dict are supported.')
 
 def normalize_segments_df(df: pd.DataFrame, segment_col: str = 'segment') -> pd.DataFrame:
```
mikkelfo, line 18: Should force correct naming of segment col

mikkelfo (changes requested):
I suggest that all our `functional` functions takes the correct inputs (i.e. not general like `concepts` and `patients_info`) and outputs the result, rather than modifying the input.

I think modifying the input and taking general arguments should be reserved for `classes`

Lastly, we should be consistent and force proper naming here and consider whether if statements are always needed or if we just force **corebehrt** features

corebehrt/common/config.py
mikkelfo: Resolved by https://github.com/FGA-DIKU/EHR/pull/5/commits/1d62fbd7a800cf5782dec9438f4f14ccb3c80994

mikkelfo (approved):
Approved with the recent changes in 1d62fbd7a800cf5782dec9438f4f14ccb3c80994, ce26e21422bf7d4cb564a0c934ce793f3d6eea10 and 08c08c37c60ed23bf369928faf9aa98a859bcdbd

DrLor (approved):

StephanLorenzen:
DrLor is my Aiomic user, which I pushed and reviewed from by error. I rolled back the commit and will commit using my regular account (this one).

StephanLorenzen (approved):

PR #7, opened 2024-07-25
Main create data fixes

StephanLorenzen:
* Added to functional:
  * split.py - functionals for splitting lists/datasets.
  * tokenize.py - functionals for tokenizing.
  * convert.py - functionals for conversion to old pipeline format. 
* Added to classes:
  * tokenize.py (moved from data and applied changes from features).
* Added to tests:
  * Test of functional.tokenize
* Updated `main_create_data.py` to use new functionals and classes, and have correct formatting.
* Updated data config (to run the script from root dir) and example data (to work with new naming).
* Updated README with details on create data step of the pipeline (mostly taken from features branch).

**Note1:** `functionals/tokenize.py` is inspired by the features-branch, but with some fixes relating to Dask and removing side-effects.
**Note2:** `main_create_data.py` is updated and works, but still needs more work, and probably should also be placed with the other main-scripts in a sub-module (e.g. like yucca/pipeline/run).


corebehrt/functional/tokenize.py
```diff
@@ -0,0 +1,100 @@
...
+    """Add separator token after each segment in the dataframe"""
+    return _add_token(features, sep_token, _get_segment_change, 1e-3)
+
+
+def add_cls_token(features: dd.DataFrame, cls_token: str = "[CLS]"):
+    """Add a classification token to the beginning of each patient's sequence"""
+    return _add_token(features, cls_token, _get_first_event, -1e-3)
+
+
+def tokenize(
```
StephanLorenzen, line 40: Come to think of it, this should probably take just the concepts Series as input instead.

mikkelfo: Agree. We should minimize side effects as much as possible

corebehrt/functional/tokenize.py
```diff
@@ -0,0 +1,100 @@
...
+        features = tokenize_frozen(features, vocabulary, inplace=inplace)
+    else:
+        features, vocabulary = tokenize_with_update(
+            features, vocabulary, inplace=inplace
+        )
+
+    return features, vocabulary
+
+
+def tokenize_with_update(
```
mikkelfo, line 58: Same here, should just take concepts rather than features

mikkelfo (changes requested):
Looks good, but lets minimize side effects before merging (take concept Series instead of features DataFrame in `tokenize.py`

corebehrt/functional/tokenize.py
DrLor: Fixed.

corebehrt/functional/tokenize.py
DrLor: Fixed.

corebehrt/functional/tokenize.py
StephanLorenzen: Fixed.

corebehrt/functional/tokenize.py
StephanLorenzen: Fixed.

StephanLorenzen:
I fixed the discussed changes:
* concepts as input (and output) instead of features dataframe for the four methods in functional/tokenize (and removed parameter inplace, as it makes less sense now).
* Updated tests for these methods.
* Updated classes/tokenize to the changed methods.

mikkelfo (approved):
Looks good. Please double-check @kirilklein

kirilklein (approved):

kirilklein (approved):

kirilklein:
Very Good!

PR #10, opened 2024-08-12
Fix pipeline

kirilklein:
The full pipeline is working now. 

Summary of the changes:
1.  Data Creation
     - Bg and death were previously removed
2. Outcome creation 
     - Minor bug fixes 
4. Finetuning
      - Fixes related to model, mostly attention mask which is now constructed on the fly in the dataset
      - trainer had to be fixed, previous changes messed up evaluation
      - fine-tuning script was restructured

### DIFF
--- a/corebehrt/classes/features.py
+++ b/corebehrt/classes/features.py
@@ -40,9 +40,11 @@ class FeatureCreator:
         )
         # Create segments
         grouped_object = concepts.groupby("PID")["ADMISSION_ID"]
+        concepts.drop("ADMISSION_ID", axis=1, inplace=True)
         concepts["segment"] = create_segments(grouped_object)
         # Create abspos
         concepts["abspos"] = create_abspos(concepts["TIMESTAMP"], self.origin_point)
+        concepts.drop("TIMESTAMP", axis=1, inplace=True)
         # Create death
         death = create_death(
             patients_info, concepts[["PID", "segment"]], self.origin_point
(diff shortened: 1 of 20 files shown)


corebehrt/classes/features.py
kirilklein, line 1: Otherwise TIMESTAMP and ADMISSION_ID values are None and the nan filter which is the next step drops death and bg

corebehrt/data/filter.py
kirilklein, line 1: We can keep it for now. Later we will have all the data in one table and filtering will be more straightforward

corebehrt/data/utils.py
kirilklein, line 1: the previos version resulted in an error

corebehrt/data_fixes/censor.py
kirilklein, line 1: Removed attention mask, since at data preparation we have the unpadded sequence and the attention mask are just ones

corebehrt/evaluation/metrics.py
kirilklein, line 1: didn't produce a full cm f targets and predictions have only one class

corebehrt/main/finetune_cv.py
kirilklein, line 1: Put code into main function. This currently requires to pass a lot of arguments around, so should be improved

corebehrt/classes/features.py
StephanLorenzen: I see. Currently we also drop them in main/create_data.py (l: 169), but after the excluder. So we should do it only one place, either here, or move it in main/create_data.py to be before the excluder.

StephanLorenzen (changes requested):
Its good!

My main confusion right now is how to actually run the pipeline. I set up the first step to be run as `python -m corebehrt.main.create_data` from the repo root dir, but this no longer works.

Can you remove it from the README, and add instructions for how to run the pipeline?

StephanLorenzen (approved):

PR #18, opened 2024-09-10
Improved outcomes

kirilklein:
Improved outcomes:

-  Tables with all outcomes and censor_outcomes for a patient which match the pattern are produced
-  Each outcome is in a different table

Outcomehandler

-  Assigning outcomes to data 
-  Filtering patients
-  Remove patients with outcomes before censor_outcomes
-  Introduce follow-up time. E.g. only consider outcomes from n hours after censor_outcome.


Montgomeryyyy (approved):

mikkelfo:
Just a minor comment: 

We can even go one step further and reduce outcomes to a single big table - rather than a dictionary of tables. It's more efficient in memory, keeps everything as dataframes, and allows us to dump a csv/parquet for _**all**_ outcomes.

kirilklein:
Yeah, but that assumes that you know all the combinations of outcome/censor outcomes from the get-go. I often run into the situation that I have an outcome which I use for censoring but then add new outcomes as targets. I think keeping it in separate tables keeps the format simple (you only need to specify the table not the columns in the table since these are always the same) and gives us more flexibility since we can load any table as censor outcome/ outcome.

mikkelfo:
Fair point, might not be worth to do it then - although we can still dump separate tables even if `data` is a dataframe instead of dictionary.

PR #22, opened 2024-09-18
Small bug fix in segments.

kirilklein:
* segments were created as n+1 for bg and then 1,1,1,2,2, 2, 3, ..., n
* now bg is assigned segment 0.
* sort by pids and timestamps in creator solves the issue


mikkelfo:
The 'expected_features` are wrong and inconsistent with your change, as they are not sorted by PID and TIMESTAMP like your commit. Simply sorting `expected_features` like your commit solves the issues, but introduces as new as `expected_features` is fundamentally wrong as the `background_gender` tokens are assigned a segment `n` rather than `0`.

PR #23, opened 2024-09-18
Seq2tab finetune

kirilklein:
- Finished conversion of prepare_finetune_data to operate on tabs.
- Incorporated Marias latest changes from the branch mlm_features.


corebehrt/functional/load.py
```diff
@@ -0,0 +1,40 @@
...
+    """Loads pids from multiple files or one file. Doesn't preserve order."""
+    if isinstance(files, str):
+        return set(torch.load(files))
+    pids = set()
+    for file in files:
+        pids.update(set(torch.load(file)))
+    return pids
+
+
+def load_predefined_pids(
```
mikkelfo, line 20: Suggest using `pathlib.Path` instead to easier check for suffixes and `is_dir` - but this is general for the entire codebase

corebehrt/functional/utils.py
```diff
@@ -57,6 +77,17 @@ def get_background_length(features: dict, vocabulary) -> int:
...
 
+def get_background_length_dd(features: dd.DataFrame, vocabulary) -> int:
+    """Get the length of the background sentence, first SEP token included."""
+    background_tokens = set([v for k, v in vocabulary.items() if k.startswith("BG_")])
+    first_pid_value = features["PID"].compute().iloc[0]
+    first_pid = features[features["PID"] == first_pid_value]
+    all_concepts_first_pid = first_pid["concept"].compute().tolist()
+    background_length = len(set(all_concepts_first_pid) & background_tokens)
+
+    return background_length + 2  # +2 for [CLS] and [SEP] tokens
```
mikkelfo, line 88: Fine for now to manually add `+2` for CLS and SEP, but we should think of a better way

corebehrt/functional/utils.py
```diff
@@ -95,7 +126,14 @@ def filter_table_by_pids(df: pd.DataFrame, pids: List[str]) -> pd.DataFrame:
     Assumes that the table has a column named PID.
     Returns a new table with only the rows that have a PID in pids
     """
-    return df[df.PID.isin(pids)]
+    return df[df.PID.isin(set(pids))]
+
+
+def exclude_pids(data: dd.DataFrame, pids_to_exclude: List[str]) -> dd.DataFrame:
```
mikkelfo, line 132: Do we want to call a function that does nothing if the 2nd argument is empty? I get this is for performance reasons, but is it our task to optimize a dumb request?

corebehrt/functional/utils.py
```diff
@@ -106,9 +144,91 @@ def remove_missing_timestamps(df: pd.DataFrame) -> pd.DataFrame:
     return df[df.TIMESTAMP.notna()]
 
 
-def get_first_event_by_pid(df: pd.DataFrame):
+def get_first_event_by_pid(df: pd.DataFrame, timestamp_column: str) -> pd.DataFrame:
```
mikkelfo, line 147: Assume `abspos` like other functions do

corebehrt/functional/utils.py
```diff
@@ -106,9 +144,91 @@ def remove_missing_timestamps(df: pd.DataFrame) -> pd.DataFrame:
...
+
+def get_gender_token(gender: str, vocabulary: dict) -> int:
+    """
+    Retrieves the token for the specified gender from the vocabulary.
+    Assumes that the gender starts with BG_GENDER_.
+    """
+    gender_key = f"BG_GENDER_{gender}"
+    try:
+        return vocabulary[gender_key]
+    except KeyError:
```
mikkelfo, line 226: Is this not the default behavior of a dictionary, except slightly more detailed? I suggest just letting the dictionary throw its standard error for clarification

corebehrt/main/create_data.py
```diff
@@ -100,29 +99,59 @@ def main_data(config_path):
...
-        feats_ft, join(cfg.output_dir, tokenized_dir_name, "features_finetune.pt")
+        df_test.compute()["PID"].unique().tolist(),
+        join(cfg.output_dir, tokenized_dir_name, "pids_test.pt"),
     )
-    torch.save(pids_ft, join(cfg.output_dir, tokenized_dir_name, "pids_finetune.pt"))
-
-    torch.save(feats_test, join(cfg.output_dir, tokenized_dir_name, "features_test.pt"))
-    torch.save(pids_test, join(cfg.output_dir, tokenized_dir_name, "pids_test.pt"))
-    logger.info("Finished tokenizing")
+    torch.save(
```
mikkelfo, line 125: Should delete old stuff

corebehrt/test.ipynb
mikkelfo, line 1: Shouldn't be part of the commit?

tests/test_main/test_create_data.py
```diff
@@ -102,11 +102,11 @@ def test_create_data(self):
...
 
-            features_path = join(self.tokenized_dir, f"features_{suffix}.pt")
-            self.assertTrue(exists(features_path))
-            features = torch.load(features_path)
-            expected_features = torch.load(
-                f"./tests/data/prepped/tokenized/features_{suffix}.pt"
-            )
-            for key in features.keys() - {"concept"}:
-                self.assertEqual(features[key], expected_features[key])
+            # features_path = join(self.tokenized_dir, f"features_{suffix}.pt")
```
mikkelfo, line 105: Remove unused stuff

mikkelfo (changes requested):
I don't know how much is overlapping with #21, but there are some minor changes and philosophical considerations to be made

corebehrt/classes/outcomes.py
```diff
@@ -131,72 +127,84 @@ def handle(
...
+        """
+        Synchronize patients in timestamps with data.
+        timestamps should be indexed by PID.
+        PIDs in timestamps that are not present in data will be assigned nan as timestamp.
+        """
+        logger.info("Synchronizing patients in data with timestamps.")
+        pids = get_pids(data)
+        timestamps = timestamps.reindex(pids)
+        timestamps = timestamps.astype(
+            pd.Int64Dtype()
```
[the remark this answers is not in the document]
kirilklein: It depends, I think an hourly resolution is sufficient in the vast majority of cases.

corebehrt/functional/data_check.py
```diff
@@ -0,0 +1,21 @@
...
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def check_max_segment(data: dd.DataFrame, vocab_size: int) -> None:
+    """Check max segment. If it's larger or equal to the model type vocab size, change accordingly."""
+    max_segment = data["segment"].max().compute()
+    if max_segment >= vocab_size:
+        logger.warning(
```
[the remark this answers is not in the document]
kirilklein: good point

corebehrt/classes/outcomes.py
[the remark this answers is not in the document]
mikkelfo: 100%, but the calculation of abspos naturally produces a float and as time2vec doesn't care, I don't see the point of forcing abspos into an integer. At best it doesn't matter and at worst we're spending more computation and limiting temporal resolution for no apparent gain imo

corebehrt/functional/filter.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def censor_data(data: dd.DataFrame, censor_dates: pd.Series) -> dd.DataFrame:
+    """
+    Censors the data by removing all events that occur after the censor_dates.
+    args:
+        data: dd.DataFrame (needs to have abspos column)
+        censor_dates: pd.Series (index: PID, values: censor_dates as abspos)
+    """
+    return filter_events_by_abspos(data, censor_dates, "<=")
```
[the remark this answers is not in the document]
kirilklein: Sure, let's just pass the callable

corebehrt/functional/utils.py
kirilklein: Well if we assume that we always have SEP tokens we could just measure the length until the first one

corebehrt/functional/utils.py
kirilklein: Ok

corebehrt/functional/utils.py
kirilklein: Yes, I think it would be a good idea to have these constants in a script and import them

corebehrt/functional/utils.py
kirilklein: Yes, perhaps finding the first SEP token and taking that as bg len?

corebehrt/functional/data_check.py
```diff
@@ -0,0 +1,21 @@
...
+            f"You've set vocab too low. Max segment {max_segment} >= type_vocab_size {vocab_size}\
+                         Change it to {max_segment+1}."
+        )
+
+
+def log_features_in_sequence(data: dict) -> None:
```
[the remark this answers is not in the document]
kirilklein: still used at this point. Will replace it with new log in the next step:)

corebehrt/functional/load.py
```diff
@@ -0,0 +1,40 @@
+""" Utils for loading data """
+
+from typing import List, Union
+import torch
+import os
+from os.path import join
+
+
+# Taken from common.loader
+def load_pids(files: Union[List, str]) -> List:
```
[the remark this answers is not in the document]
kirilklein: Yeah, probably sufficient for most cases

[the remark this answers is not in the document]
kirilklein: Let's adress that in a separate issue

corebehrt/functional/load.py
kirilklein: Same here

mikkelfo (approved):
Looks good to be, but I also lost the overview a slight bit. Should be good to merge

PR #39, opened 2024-10-11
Restructure common utils

kirilklein:
Add functions from common.utils to modules inside functional.
Data class is still used but will be removed shortly, so we leave it for now in common. utils.
Also, functions for conversion from epochs to num_steps were refactored.

### DIFF
--- a/corebehrt/common/utils.py
+++ b/corebehrt/common/utils.py
@@ -5,7 +5,7 @@ import random
 from copy import deepcopy
 from dataclasses import dataclass, field
 from os.path import join
-from typing import Dict, Generator, List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import pandas as pd
 import numpy as np
@@ -15,22 +15,6 @@ from corebehrt.common.config import Config
 
 logger = logging.getLogger(__name__)  # Get the logger for this module
 
-def iter_patients(features: dict) -> Generator[dict, None, None]:
-    """Iterate over patients in a features dict."""
-    for i in range(len(features["concept"])):
-        yield {key: values[i] for key, values in features.items()}
-
-def check_existing_splits(data_dir: str)-> bool:
-    """Check if train, val, and test splits already exist in data_dir."""
-    if os.path.exists(join(data_dir, 'train_pids.pt')) and\
-        os.path.exists(join(data_dir, 'val_pids.pt')) and\
-        os.path.exists(join(data_dir, 'test_pids.pt')) and\
-        os.path.exists(join(data_dir, 'train_file_ids.pt')) and\
-        os.path.exists(join(data_dir, 'val_file_ids.pt')) and\
-        os.path.exists(join(data_dir, 'test_file_ids.pt')):
-        return True
-    else:
-        return False
     
 def split_path(path_str: str) -> list:
     """Split path into its components."""
--- a/corebehrt/common/utils.py
+++ b/corebehrt/common/utils.py
@@ -20,12 +20,6 @@ def iter_patients(features: dict) -> Generator[dict, None, None]:
     for i in range(len(features["concept"])):
         yield {key: values[i] for key, values in features.items()}
 
-def check_patient_counts(concepts: pd.DataFrame, patients_info: pd.DataFrame, logger)->None:
-    """Check that the number of patients in concepts and patients_info match."""
-    if concepts.PID.nunique() != patients_info.PID.nunique():
-            logger.warning(f"patients info contains {patients_info.PID.nunique()} patients != \
-                        {concepts.PID.nunique()} unique patients in concepts")
-
 def check_existing_splits(data_dir: str)-> bool:
     """Check if train, val, and test splits already exist in data_dir."""
     if os.path.exists(join(data_dir, 'train_pids.pt')) and\
--- a/corebehrt/common/utils.py
+++ b/corebehrt/common/utils.py
@@ -1,4 +1,3 @@
-import glob
 import logging
 import os
 import random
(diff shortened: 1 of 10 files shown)


corebehrt/common/utils.py
```diff
@@ -74,23 +21,29 @@ class Data:
...
     vocabulary: Optional[Dict] = field(default=None)
     mode: Optional[str] = field(default=None)
-    
+
     def __len__(self):
         return len(self.pids)
-    
-    def copy(self) -> 'Data':
+
+    def copy(self) -> "Data":
```
mikkelfo, line 28: I think we should be able to just `deepcopy(self)` and all attributes should automatically be deepcopied as well

corebehrt/common/utils.py
```diff
@@ -100,80 +53,109 @@ def load_tensor(filename, required=False):
                 else:
                     return None
             return torch.load(filepath)
-        features = load_tensor(f'{mode}_features.pt', required=True)
-        pids = load_tensor(f'{mode}_pids.pt', required=True)
-        outcomes = load_tensor(f'{mode}_outcomes.pt')
-        censor_outcomes = load_tensor(f'{mode}_censor_outcomes.pt')
-        vocabulary = load_tensor('vocabulary.pt')
+
+        features = load_tensor(f"{mode}_features.pt", required=True)
```
mikkelfo, line 57: I think we should check if filepath exists and load if exists (using `path().exists()`) rather than "loading" a file that isn't present for clarity

corebehrt/common/utils.py
```diff
@@ -100,80 +53,109 @@ def load_tensor(filename, required=False):
...
-        return Data(features={key: [values[i] for i in indices] for key, values in self.features.items()}, 
-                        pids=[self.pids[i] for i in indices],
-                        outcomes=[self.outcomes[i] for i in indices] if self.outcomes is not None else None,
-                        censor_outcomes=[self.censor_outcomes[i] for i in indices] if self.censor_outcomes is not None else None,
-                        vocabulary=self.vocabulary,
-                        mode=mode)
-    
-    def select_data_subset_by_pids(self, pids: list, mode:str='')->'Data':
+
+    def select_data_subset_by_indices(self, indices: list, mode: str = "") -> "Data":
```
mikkelfo, line 87: This should maybe be a part of the `Data` class as a `subset` method, as it only functions on this

corebehrt/functional/data_check.py
```diff
@@ -19,3 +21,17 @@ def log_features_in_sequence(data: dict) -> None:
     logger.info("Example features: ")
     for k, v in data.features.items():
         logger.info(f"{k}: {v[0]}")
+
+
+def check_patient_counts(
```
mikkelfo, line 26: Should this only issue a warning or is this an error?

corebehrt/functional/files_and_folders.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+
+def split_path(path_str: str) -> list:
```
mikkelfo, line 4: This can be simplified using `pathlib Path` by doing `Path(path_str).parts` to split up and extract directories (by doing `parts[:-1]`

corebehrt/main/pretrain.py
```diff
@@ -38,8 +38,10 @@ def main_train(config_path):
     train_dataset, val_dataset = DatasetPreparer(cfg).prepare_mlm_dataset()
 
     if "scheduler" in cfg:
```
mikkelfo, line 40: This is duplicated for finetune, so perhaps it should be wrapped in a functionality for standardization?

corebehrt/main/finetune_cv.py
```diff
@@ -133,8 +134,11 @@ def finetune_fold(
 ) -> None:
     """Finetune model on one fold"""
     if "scheduler" in cfg:
```
mikkelfo, line 136: See comment below for `pretrain.py`

mikkelfo (changes requested):
Feel free to ignore my comments for the `Data` dataclass if it is getting removed, just thought I'd add my comments to the record

corebehrt/common/utils.py
kirilklein: Do you mean doing a check inside the load_from_directory method instead of inside the submethod? 
Since not all the files need to be present to construct the data, we check if optional files are in the directory and return None if not. What would you suggest instead?

We might figure out a better way to do it for the new dataclass.

corebehrt/common/utils.py
kirilklein: We won't need the method in the future

kirilklein:
> Feel free to ignore my comments for the `Data` dataclass if it is getting removed, just thought I'd add my comments to the record

OK. I don't think we should completely get rid of it, since a dataclass also makes sense with dataframes, especially if we have outcomes and censor_outcomes in separate dfs. This will help us keep track of the patients and allow for some checks. However, most methods won't be needed in the new setup.

corebehrt/common/utils.py
kirilklein: Good point! We should note that for the df version of the method.

corebehrt/functional/data_check.py
kirilklein: I think this method is no longer needed. Also, we'll get rid of the ConceptLoaderLarge.
It's probably a good idea to tackle create_outcomes next to rewrite with dask

kirilklein: Adressed in a8129132b34860ad4e39afbd1131375127d61d3f

corebehrt/functional/files_and_folders.py
kirilklein: Addressed in 52be7d6314f3ad08240774c9eefd34fa17927768

corebehrt/main/pretrain.py
kirilklein: I'm not sure what you mean exactly. You mean the whole thing with the if statement?

kirilklein:
@mikkelfo I've adressed your comments, except the ones for the dataclass as we'll rewrite it and get rid of most methods.
Could you please clarify this comment;
> This is duplicated for finetune, so perhaps it should be wrapped in a functionality for standardization?

corebehrt/common/utils.py
mikkelfo: Yeah there might be a few options to "fix" it IMO. Just feel like it's a little odd to have a bunch of load statements that might be doing
We might do it inside the load_from_directory method or maybe we can just change the `load_tensor` to `load_if_exists` or something like that to clarify. Or have a pretrain/finetune one, so we only load the expected files or something like that

corebehrt/main/pretrain.py
mikkelfo: Yes. If we have to adjust the scheduler in the same way for both pretraining and finetune, we should standardize it, rather than having it duplicates in both scripts

mikkelfo: But now where I look at it today, I guess it's just an if_statement with a log, so might not be needed

mikkelfo (approved):
I don't have direct solutions to my other comments (and they might not be valid), so it's fine to merge now imo

PR #44, opened 2024-10-19
Convert targets to numpy before saving

### LINKED ISSUE #43: Targets are not saved correctly when using numpy.savez_compressed
https://github.com/FGA-DIKU/EHR/blob/5f0a18aca2e86b2a0b8f04634ce105da27db608f/corebehrt/trainer/utils.py#L43

kirilklein:
close #43


mikkelfo (approved):
Should just force-merge these small pull requests instead of waiting for approval imo

PR #45, opened 2024-10-21
Move model

kirilklein:
* Move model, activations, heads and embeddings to classes 
* Pass args explicitly instead of passing config to heads, activations and embeddings


corebehrt/classes/embeddings.py
```diff
@@ -0,0 +1,120 @@
...
+        hidden_dropout_prob: float,
+    ):
+        super().__init__()
+        self.LayerNorm = nn.LayerNorm(hidden_size, eps=layer_norm_eps)
+        self.dropout = nn.Dropout(hidden_dropout_prob)
+
+        # Initalize embeddings
+        self.concept_embeddings = nn.Embedding(vocab_size, hidden_size)
+        self.age_embeddings = Time2Vec(
+            1,
```
mikkelfo, line 46: This is always going to be `1` right? If so, perhaps we should just remove the `input_dim` argument for Time2Vec to simplify, as it doesn't make too much sense

corebehrt/classes/embeddings.py
```diff
@@ -0,0 +1,120 @@
...
+        embeddings = self.LayerNorm(embeddings)
+        embeddings = self.dropout(embeddings)
+
+        return embeddings
+
+
+class Time2Vec(torch.nn.Module):
+    def __init__(
+        self,
+        input_dim: int = 1,
```
mikkelfo, line 87: See above

corebehrt/classes/heads.py
```diff
@@ -2,16 +2,16 @@
...
 class MLMHead(torch.nn.Module):
-    def __init__(self, config):
+    def __init__(self, hidden_size: int, vocab_size: int, layer_norm_eps: float):
         super().__init__()
         # BertPredictionHeadTransform
-        self.dense = torch.nn.Linear(config.hidden_size, config.hidden_size)
+        self.dense = torch.nn.Linear(hidden_size, hidden_size)
         self.activation = torch.nn.GELU()
-        self.LayerNorm = torch.nn.LayerNorm(config.hidden_size, eps=config.layer_norm_eps)
+        self.LayerNorm = torch.nn.LayerNorm(hidden_size, eps=layer_norm_eps)
```
mikkelfo, line 10: Should probably keep this to snake_case instead of CamelCase (I can see github "sees" it as a Class instead of an attribute)

mikkelfo (approved):
Minor semantic changes can be done, otherwise fine

PR #53, opened 2024-10-29
DirectoryPreparer and Azure to SDK v2

StephanLorenzen:
Changes:
* Config-files: 
  * Made the configuration of paths more consistent using `paths` sub-config. Input/output paths are now all specified in `paths`. Either full paths must be set (or `paths.root` can be set, and default paths will be generated).
  * Added `logging` sub-config for greater log control.
* Updated  DirectoryPreparer which will now
  * Make sure that required paths in the config are set (thus removing the need for a lot of checks later, where paths can be assumed to be set),
  * Checks that required dirs are present and creates output dirs.
* Moved some code from Initializer and main scripts to DirectoryPreparer (e.g. path validation and cfg file copying).
* Removed all Azure SDK v1 related code.
* Added new Azure submodule `corebehrt.azure` with wrappers for running the main scripts in Azure SDK v2.
* Added the full corebehrt dir to the Lint workflow (as it was almost already passing the check).
* Added additional unit tests of the main scripts:
  * Create outcomes
  * Pretrain
  * Pretrain from restarted model (although I do not know of a good way to actually detect whether the model is updated).

TODO:
* Logging to Azure metrics.
* I think we should expand ModelManager to also be useable in the pretrain setup, or possible have two ModelManager classes with one inheriting the other (currently we are doing a lot of the same in main/pretrain but without the Manager). It might also make sense to incorporate Initializer, as there also seems to be some overlap.


corebehrt/azure/__main__.py
```diff
@@ -0,0 +1,91 @@
...
+    # Job parser
+    job_parser = subparsers.add_parser("job", help="Run a single job.")
+    job_parser.add_argument(
+        "JOB",
+        type=str,
+        choices=("create_data", "create_outcomes", "pretrain", "finetune_cv"),
+        help="Job to run.",
+    )
+    job_parser.add_argument(
+        "-c",
```
kirilklein, line 54: Can we change the flage to something else to not introduce conflicts with the -c compute flag?

corebehrt/azure/__main__.py
kirilklein, line 1: I would break it further refactor it into functions e.g.
```
    if args.call_type == "job":
        run_job(args, register_output)
    else:
        parser.print_help()
        sys.exit(1)
```
And basically, everything that is a comment followed by a code block can just be a function instead:)

corebehrt/azure/util.py
kirilklein, line 1: Can we break up setup_job into smaller, more readable functions?

corebehrt/common/setup.py
```diff
@@ -27,191 +32,360 @@ def get_args(default_config_name, default_run_name=None):
...
-        finetune_folder_name = f"{finetune_folder_name}_followup_start_{n_hours_follow_up_str}_index_date_{cfg.paths.run_name}"
-        return finetune_folder_name
+        return f"{run_id}_followup_start_{n_hours_follow_up_str}_index_date_{suffix}"
 
     @staticmethod
     def get_event_name(path: str) -> str:
+        """
+        Gets the event name from the path to the outcome file.
+        """
         return split(path)[-1].strip(".csv")
```
kirilklein, line 389: It'd be better to use splitext from os.path

corebehrt/common/setup.py
```diff
@@ -27,191 +32,360 @@ def get_args(default_config_name, default_run_name=None):
...
-                "Either pretrain_model_path, model_path, or save_folder_path must be provided."
-            )
-        cfg.paths.output_path = output_path
-        cfg.paths.run_name = DirectoryPreparer.construct_finetune_model_dir_name(cfg)
-        return cfg
+        return uuid.uuid4().hex
 
-    @staticmethod
-    def construct_finetune_model_dir_name(cfg: Config) -> str:
+    def generate_finetune_run_id(self) -> str:
```
kirilklein, line 345: We could also refactor this function and break down into smaller helper functions.
Also, I think the previous name was more descriptive

kirilklein (changes requested):
Very good, I just had a few minor suggestions regarding refactoring functions to improve readability and maintainability.

corebehrt/main/pretrain.py
mikkelfo, line 1: Is restarting only implemented for pretraining?

tests/data/pretrain/pretrain_config.yaml
```diff
@@ -0,0 +1,55 @@
...
+    _target_: corebehrt.evaluation.metrics.PrecisionAtK
+    topk: 10
+model:
+  abspos_embeddings: false
+  hidden_size: 96
+  intermediate_size: 64
+  linear: true
+  num_attention_heads: 3
+  num_hidden_layers: 3
+  plusplus: true
```
mikkelfo, line 32: Some of these things (`plusplus`, `linear` and `abspos_embeddings`) has been made default in the model. Lets just remove them here

tests/test_main/test_pretrain.py
```diff
@@ -0,0 +1,154 @@
...
+from .base import TestMainScript
+
+
+class TestCreateOutcomes(TestMainScript):
+    def setUp(self):
+        super().setUp()
+
+        self.pretrain_dir = join(self.tmp_dir, "pretrain")
+
+        self.set_config(
```
mikkelfo, line 15: Shouldn't we be using the default test configs, instead of having a hardcoded one?

mikkelfo (changes requested):
Mostly minor stuff that can be addressed later (but I thought we might as well do it now). Main thing is restarting is only implemented for pretraining and the test configs are not (?) being used in the tests

corebehrt/common/setup.py
mikkelfo: I'm fine with having it in a single function, as this is _only_ relevant for the naming of the finetune folder. For naming, I can see an argument for both, so don't really care

corebehrt/common/setup.py
mikkelfo: Going forward, I would prefer using `pathlib` rather than strings in general. Could leave this for future PR

corebehrt/azure/__main__.py
DrLor: I thought about this, it should not cause a conflict as it is for the sub command - but I think the change is good for clarity anyways. I will change it

StephanLorenzen: I thought about this, it should not cause a conflict as it is for the sub command - but I think the change is good for clarity anyways. I will change it

corebehrt/common/setup.py
StephanLorenzen: I thought about this, it should not cause a conflict as it is for the sub command - but I think the change is good for clarity anyways. I will change it

corebehrt/common/setup.py
StephanLorenzen: I think I had a reason for using run_id instead of dir_name (maybe azure related), but I do not remember, so it probably ended up not being relevant :-)
Fine to use dir name for me.

Wrt. function splitting - here and in general - I must admit I rarely split (shorter) single-use functions, if I do not think it improves the readability. "shorter" is of course very subjective. Maybe we need a coding guideline on this, so as to not have it as a comment for every PR :-)

corebehrt/main/pretrain.py
StephanLorenzen: No, it is for finetune as well. For finetune, it is handled in ModelManager, which handles finetune models, in contrast to pretrain, where it is in the main script.

Hence my suggesting in the PR to extend ModelManager (alt. inherit) to also be usable for pretrain, so we have a consistent model management across the two scripts. But as the PR is already quite large, I did not want to make that change now :-)

Another thing I forgot to say - it also supports both restarting from `paths.model` (if passing an existing model) or from `paths.restart_model`. The first for backwards compatibility (as that functionality was currently there), the second because Azure needs separate inputs and outputs (I think...).

tests/data/pretrain/pretrain_config.yaml
StephanLorenzen: Makes sense, but could probably be done as part of the config documentation issue #46 (as we have to look through the configs then anyways)

tests/test_main/test_pretrain.py
StephanLorenzen: The initial idea with hard-coding the config was to make it easy to change/re-create between tests. We also did not have test config files in the `tests` directory then.

I had to add them to the `tests/data/{features,tokenized,outcomes}` with the new `test_main/test_*`  tests, and if we are not changing them a lot, we can read them from there instead now :-)

corebehrt/common/setup.py
StephanLorenzen: I agree on both (as I do not want to go through all paths with pathlib in this PR).

corebehrt/data/prepare_data.py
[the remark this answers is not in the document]
DrLor: I changed `paths` back to `paths_cfg` in `loader`.

corebehrt/azure/util.py
```diff
@@ -0,0 +1,211 @@
...
+    :param inputs: argument configuration for input directories, i.e. a mapping
+        to configuration elements, typically in the paths sub-config.
+    :param outputs: argument configuration for output directories.
+    :param config: The config for the current job.
+    :param compute: The azure compute to use for the job.
+    :register_output: A mapping from output id to name, if the output should be
+        registered as a data asset.
+    """
+    check_azure()
+    assert compute is not None
```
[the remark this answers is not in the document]
DrLor: Removed, also made it non-optional in the command line utility.

corebehrt/common/setup.py
DrLor: I renamed it (and renamed `run_id` => `run_name` as well.
I did not break it further down, except the suffix, where we already had another utility function, which did exactly the same.

corebehrt/azure/util.py
DrLor: It is done.

corebehrt/azure/__main__.py
DrLor: Also done.

corebehrt/azure/__main__.py
DrLor: I changed compute to be non-optional instead, so no conflicts.

StephanLorenzen: I changed compute to be non-optional instead, so no conflicts.

corebehrt/azure/__main__.py
StephanLorenzen: Also done.

corebehrt/azure/util.py
StephanLorenzen: It is done.

corebehrt/common/setup.py
StephanLorenzen: I renamed it (and renamed `run_id` => `run_name` as well).
I did not break it further down, except the suffix, where we already had another utility function, which did exactly the same.

corebehrt/azure/util.py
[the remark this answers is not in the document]
StephanLorenzen: Removed, also made it non-optional in the command line utility.

corebehrt/data/prepare_data.py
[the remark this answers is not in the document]
StephanLorenzen: I changed `paths` back to `paths_cfg` in `loader`.

corebehrt/azure/__main__.py
kirilklein: ok

kirilklein (approved):

tests/test_main/test_pretrain.py
mikkelfo: I think we should, be we can leave that for future as well

mikkelfo (approved):

PR #56, opened 2024-11-17
Create data saves parquet

kirilklein:


kirilklein:
Use parquet format for saving features and tokenized features. 
Update test and test data.

mikkelfo (approved):
Looks good to me. 

(We can consider (for a future PR) to also convert concepts files into parquet)

PR #64, opened 2024-11-25
Speed up/ fix tokenization

kirilklein:
* Combine sep and cls token addition into one func to speed up
* get rid of unnecessary sorts
* broadcast tokenize with update
* use map partitions for special token addition in tokenizer
* set index right from the start !todo: sped up frozen tokenization by getting rid of lambda function


kirilklein:
### time:

-  1M tokenization PT  (no ading of special tokens): 950s

kirilklein:
@Montgomeryyyy please make sure the sorting also works for lab tests and values

Issue #67, opened 2024-11-28
Environment: Specify python version as <3.13


kirilklein:
There are compatibility issues of python 3.13 and torch.save
see [this issue](https://github.com/python/cpython/issues/125631)

StephanLorenzen:
Closed with #68

PR #68, opened 2024-12-02
Azure env builder

StephanLorenzen:
* Updated versions of libraries in our requirements.
* Added utility for building the Azure environment CoreBEHRT from the current requirements.txt with the correct python version.
* Added weights_only to torch.load(...) calls to get rid of torch warning.
* pids are now saved as lists instead of sets (needed for weights_only). 
* Slightly changed the Azure command line utility.


StephanLorenzen:
Closes #66 and #67

kirilklein (approved):
Looks good

PR #79, opened 2025-01-10
Revamp of create_data that handles RESULTs as a list

mikkelfo:
Took some liberties w.r.t. some simplifications (mainly related to Values). Admittedly I lost a little overview due to a lot of dask bugs and weird behaviour, so there might be a few fixes to do 

### Changelog summary

FeatureCreator:
	Treats CONCEPT and RESULT as a ‘concept’ list column for easier ordering
	Added CLS creation as part of create_background (CLS is now a part of background)
	Added SEP creation to the last segment unique segment of each patient
	Made min-max-normalization default on ‘RESULT’ column

Deleted FormattedDataLoader
Simplified ValuesNormalizer and made min-max-normalization default 
Deleted ValueCreator and made it part of FeatureCreator 
Updated configs due to special token creation and values simplification 
Removed redundant exclusion function
Simplified functional min_max_normalization - now handles single-values RESULT 
Removed special_token_assignment from tokenizer
Used functional.constants whenever possible
Adjusted tests

Creator:
	Returned feature result as Series, instead of assigning it to a column, whenever possible
	- In some places it wasn’t possible due to some weird dask interactions


tests/test_functional/test_tokenize.py
kirilklein, line 1: Is that tested in a different place now?

corebehrt/main/create_data.py
```diff
@@ -38,80 +29,97 @@
...
+
+    # Load formatted data and create background
+    raw_dir = Path(cfg.paths.data)
+    concepts = dd.concat(
+        [
+            dd.read_csv(raw_dir / f"concept.{concept_type}.csv", dtype=default_dtypes)
+            for concept_type in cfg.loader.concept_types
+        ]
+    )
+    # parse_date did not play nice, so we do conversion manually
```
kirilklein, line 51: we should extract this into a small helper function. Also, why are we using .astype here but not for birthdate and deathdate?

tests/data/features/data_config.yaml
```diff
@@ -9,21 +9,24 @@ features:
     day: 26
     month: 1
     year: 2020
+  cls_token: true
```
kirilklein, line 12: same thing here regarding cls token

kirilklein (changes requested):
I have rather minor comments, most of them we can address later and open issues for now. Most importantly, we should test on azure before merging. Also, we should profile what blows up the disk memory usage (if it's still the case, which I assume it is)

corebehrt/classes/features.py
```diff
@@ -42,36 +47,66 @@ def __init__(
...
 
         features = create_segments(features)
+
         features = features.drop(columns=["ADMISSION_ID", "TIMESTAMP", "BIRTHDATE"])
 
+        if self.sep_token:
+            features = create_sep_tokens(features)
+
+        # Explode "concept" column (list of concepts) into separate rows
+        features = features.explode("concept")
```
Montgomeryyyy, line 106: I thought we wanted to keep everything in the lists when tokenising as well as in the prepare data functions, so that it was easier to keep track of when censoring?

Montgomeryyyy (commented):
I don't really have any comments, as I feel like we discussed most of the things:)
I've lost a bit track of all the extra stuff I had, and why, so it probably makes sense to test it on the labs data, to see if it still performs similarly. 
The only thing is that I thought that we wanted to keep the concepts in list, until right at the end, so that we could also utilise the lists when censoring, such that grouped concepts stay together. Or will we just do that later?

corebehrt/configs/create_data.yaml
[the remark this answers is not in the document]
mikkelfo: I'd just set the option to `False` and leave the current functionality like it is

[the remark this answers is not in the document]
mikkelfo: I'm a bit unsure what you mean by mapping? Do you mean creating a "fake" vocab, where a long (previous cutoff) code, like `DC12345` is simply assigned the same integer as `DC12` in the vocab? 

This would be a way to keep the features the identical across cutoff option and thereby only change the tokenized versions

corebehrt/classes/tokenizer.py
```diff
@@ -56,31 +51,22 @@ def __call__(self, features: dd.DataFrame) -> dd.DataFrame:
...
 
-        return features.map_partitions(_process_partition)
+        features["concept"] = features["concept"].map_partitions(_tokenize_partition)
+        return features
 
     def update_vocabulary(self, concepts: dd.Series) -> None:
         """Create or update vocabulary from unique concepts"""
         # Get unique concepts across all partitions
-        unique_concepts = concepts.drop_duplicates().compute()
+        unique_concepts = concepts.unique().compute()
```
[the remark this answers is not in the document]
mikkelfo: I'm fine reverting this change, seems the consensus is that `drop_duplicates` is probably better and more stable

tests/test_classes/test_tokenizer.py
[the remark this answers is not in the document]
mikkelfo: SEP tokens are now interested during FeatureCreator, happens [here](https://github.com/FGA-DIKU/EHR/pull/79/files/06e71a2a108ec962719cb5e4d885e0fabeeefb8f#diff-e8974e0b9a47977c879630df61f65358e4891aa5d39a50bb586b93644c6f9e27R151) and there's also a `no_sep` test inside it

tests/test_functional/test_tokenize.py
mikkelfo: See https://github.com/FGA-DIKU/EHR/pull/79/files/06e71a2a108ec962719cb5e4d885e0fabeeefb8f#diff-e8974e0b9a47977c879630df61f65358e4891aa5d39a50bb586b93644c6f9e27

corebehrt/main/create_data.py
mikkelfo: The astype is most likely a relic so deleting it, moving the date conversion to a helper

corebehrt/classes/features.py
mikkelfo: You are completely right, but I'm scared this might be a larger revamp. I think we tackle this PR first just to get create_data working with full data and then I'll open an issue about it

corebehrt/classes/tokenizer.py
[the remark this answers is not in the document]
mikkelfo: Resolved in 53ad631d40d354f68e7a6d2fb6edbd229bdc414b

corebehrt/main/create_data.py
mikkelfo: Resolved in 9c67261c0ed2983b0c3859af25165a2d3e26fb7c

corebehrt/configs/create_data.yaml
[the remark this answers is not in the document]
mikkelfo: Done e97622d621cbd8dc83a156e4578384d12804c781

tests/data/features/data_config.yaml
mikkelfo: Done in e97622d621cbd8dc83a156e4578384d12804c781

corebehrt/classes/features.py
mikkelfo: Opened in #80 if you're fine with waiting for it

corebehrt/configs/create_data.yaml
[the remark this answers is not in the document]
kirilklein: Actually, an optional custom vocab  is the best solution, and I think we already allow for that. THen we don't need any remapping or modification inside the pipeline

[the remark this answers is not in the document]
kirilklein: So I think as a conclusion, let's just remove the cutoff options

corebehrt/classes/features.py
kirilklein: I also think we can do it in a separate PR as it requires a bit more work

kirilklein:
Works on azure (with med and diag) and is even a bit faster than the previous version. We still run out of disk with lab tests. As a temp. solution I will try to run it on a bigger cluster or alternatively set the tmp dask directory to the blobstore

corebehrt/classes/features.py
Montgomeryyyy: alright, sounds good

kirilklein:
The script throws an error when loading on azure:
```
Traceback (most recent call last):
  File "parsers.pyx", line 1161, in pandas._libs.parsers.TextReader._convert_tokens
TypeError: Cannot cast array data from dtype('O') to dtype('float64') according to the rule 'safe'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask/backends.py", line 140, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask/dataframe/io/csv.py", line 771, in read
    return read_pandas(
           ^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask/dataframe/io/csv.py", line 640, in read_pandas
    head = reader(BytesIO(b_sample), nrows=sample_rows, **head_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 1026, in read_csv
    return _read(filepath_or_buffer, kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 626, in _read
    return parser.read(nrows)
           ^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 1923, in read
    ) = self._engine.read(  # type: ignore[attr-defined]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/pandas/io/parsers/c_parser_wrapper.py", line 234, in read
    chunks = self._reader.read_low_memory(nrows)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "parsers.pyx", line 850, in pandas._libs.parsers.TextReader.read_low_memory
  File "parsers.pyx", line 921, in pandas._libs.parsers.TextReader._read_rows
  File "parsers.pyx", line 1066, in pandas._libs.parsers.TextReader._convert_column_data
  File "parsers.pyx", line 1167, in pandas._libs.parsers.TextReader._convert_tokens
ValueError: could not convert string to float: 'Kommentar   '

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask/backends.py", line 140, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask_expr/_collection.py", line 5124, in read_csv
    return new_collection(
           ^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask_expr/_collection.py", line 4803, in new_collection
    meta = expr._meta
           ^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/functools.py", line 993, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask_expr/io/csv.py", line 95, in _meta
    return self._ddf._meta
           ^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/functools.py", line 993, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask_expr/io/csv.py", line 85, in _ddf
    return self.operation(
           ^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask/backends.py", line 151, in wrapper
    raise exc from e
ValueError: An error occurred while calling the read_csv method registered to the pandas backend.
Original Message: could not convert string to float: 'Kommentar   '

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/mnt/azureml/cr/j/81481d0b0eb94ff9aed8d13d03a28ecf/exe/wd/corebehrt/azure/components/create_data.py", line 26, in <module>
    create_data.main_data(util.AZURE_CONFIG_FILE)
  File "/mnt/azureml/cr/j/81481d0b0eb94ff9aed8d13d03a28ecf/exe/wd/corebehrt/main/create_data.py", line 47, in main_data
    dd.read_csv(raw_dir / f"concept.{concept_type}.csv", dtype=default_dtypes)
  File "/azureml-envs/azureml_d5faf50b50787e45031334853f7270a3/lib/python3.12/site-packages/dask/backends.py", line 151, in wrapper
    raise exc from e
ValueError: An error occurred while calling the read_csv method registered to the pandas backend.
Original Message: An error occurred while calling the read_csv method registered to the pandas backend.
Original Message: could not convert string to float: 'Kommentar   '
```

kirilklein:
Also, we should make sure we can read both .csv and .parquet processed data. Perhaps it would be good to have a helper function for loading (and since FormattedDataLoader worked fine that would be a good place to start)

kirilklein:
> The create_data script throws an error during loading on azure:
> 
> ```
> Traceback (most recent call last):
...
> ```

I assume this is caused by labtest values which can be a str

corebehrt/configs/create_data.yaml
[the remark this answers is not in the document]
mikkelfo: Done in c646a0243efc2befa119ac7c60abc0c04e6c7531

mikkelfo:
@kirilklein Dtypes issues should be solved in 1aea5b9256545fab596ce24a16f87c5eeca230c5

Have updated RESULT to be a string column and modified `min_max_normalization` to only be calculated on numeric values. Test data is updated with a singular string value "Kommentar     " to better model azure data

mikkelfo:
> Also, we should make sure we can read both .csv and .parquet processed data. Perhaps it would be good to have a helper function for loading (and since FormattedDataLoader worked fine that would be a good place to start)

@kirilklein We can easily support both, but I think it'd be wiser to simply force a parquet format. It's **very** easy reformat  csv files to parquet and parquet should be, IMO, the standard format due to the compression. The only argument for csv is human readability, but we never open the csv files regardless, so I'd vote for just forcing a parquet format

kirilklein:
> > Also, we should make sure we can read both .csv and .parquet processed data. Perhaps it would be good to have a helper function for loading (and since FormattedDataLoader worked fine that would be a good place to start)
> 
> @kirilklein We can easily support both, but I think it'd be wiser to simply force a parquet format. It's **very** easy reformat csv files to parquet and parquet should be, IMO, the standard format due to the compression. The only argument for csv is human readability, but we never open the csv files regardless, so I'd vote for just forcing a parquet format

Agree, let's force parquet format

mikkelfo:
Left PR open until performance issues have been resolved

mikkelfo:
Closed due to #109

PR #89, opened 2025-01-26
Patient selection

kirilklein:
Add Cohort selection script including:

 * Filter by age and sex
 * Removal of dead patients
 * Removal of patients with outcomes before index date

Further changes:
* Modify prepare data to load cohort
* Simplify outcome handling in prepare data

Implement unittests for all new functionality


corebehrt/classes/patient_filter.py
```diff
@@ -0,0 +1,129 @@
...
+    if max_age is not None:
+        patients_info = patients_info[age_in_years <= max_age]
+
+    return patients_info
+
+
+def apply_exclusion_filters(
+    patients_info: pd.DataFrame,
+    outcomes: pd.DataFrame,
+    dead_before_index_date: bool = False,
```
mikkelfo, line 86: Shouldn't we set this to True as default?

corebehrt/classes/patient_filter.py
```diff
@@ -0,0 +1,129 @@
...
+        patients_info = patients_info[
+            patients_info[DEATHDATE_COL].isna()
+            | (patients_info[DEATHDATE_COL] > patients_info[TIMESTAMP_COL])
+        ]
+
+    if outcome_before_index_date and not outcomes.empty:
+        earliest_outcomes = outcomes.groupby(PID_COL)[TIMESTAMP_COL].min()
+
+        # Align earliest_outcomes with the patients we still have
+        earliest_outcomes = earliest_outcomes.reindex(patients_info[PID_COL].unique())
```
mikkelfo, line 108: I'm unsure about using `reindex`. Does this just function as a `PID.isin`, but for the index?

corebehrt/classes/patient_filter.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Filter them out
+        patients_info = patients_info[~patients_info[PID_COL].isin(pids_to_exclude)]
+
+    return patients_info
+
+
+def filter_df_by_pids(df: pd.DataFrame, pids: Set[str]) -> pd.DataFrame:
+    """Filter using PID column in dataframe."""
+    pids = set(pids)
```
mikkelfo, line 128: converting to set is slower (and redundant) as pandas internally converts it to a list or np.array

corebehrt/classes/prepare_data.py
```diff
@@ -36,14 +46,31 @@ def prepare_finetune_data(self) -> PatientDataset:
...
                 join(
                     paths_cfg.tokenized,
                     "features_finetune",
                 )
-            ).compute()
+            )
+            if pids is not None:
+                df = filter_df_by_pids(df, pids)
+            # !TODO: if index date is the same for all patients, then we can censor here.
+            df = df.compute()
```
mikkelfo, line 73: We load in the entire finetuned tokenized dataset here? That's not good

corebehrt/classes/prepare_data.py
```diff
@@ -118,14 +145,18 @@ def prepare_pretrain_data(self) -> Tuple[PatientDataset, dict]:
...
             df = dd.read_parquet(
                 join(
                     paths_cfg.tokenized,
                     "features_pretrain",
                 )
-            ).compute()
+            )
+            if pids is not None:
+                df = filter_df_by_pids(df, pids)
+            df = df.compute()
```
mikkelfo, line 159: Same as above, but even worse since the pretrain dataset is often larger than finetune

corebehrt/functional/filter.py
```diff
@@ -52,3 +52,29 @@ def filter_events_by_abspos(
...
+
+    Returns:
+        Filtered DataFrame.
+    """
+    if include_values is not None and exclude_values is not None:
+        raise ValueError(
+            f"Cannot include and exclude values for column '{column}' simultaneously."
+        )
+
+    df = df.copy()
```
mikkelfo, line 75: Is the copy needed here? A boolean isin operation doesn't do anything in place to a dataframe

corebehrt/functional/outcomes.py
```diff
@@ -0,0 +1,50 @@
...
+    Args:
+        index_dates: DataFrame with PID_COL and abspos columns
+        outcomes: DataFrame with PID_COL and abspos columns
+        n_hours_start_follow_up: Hours after index date to start follow-up
+        n_hours_end_follow_up: Hours after index date to end follow-up (None for no end)
+
+    Returns:
+        Series with PID index and int (0 or 1) values indicating if outcome occurred in window
+    """
+    # Convert hours to integer positions (assuming hourly data)
```
mikkelfo, line 22: Comment does not make sense and start_pos and end_pos is also redundant, I think we should just keeps the original args name (since we only use start_pos and end_pos 1-2 times)

mikkelfo (changes requested):
Using PIDs for index is a good idea, but also susceptible to order-related issues (e.g. when creating a new Series). I haven't found any instances of concern, it's just something we need to keep in mind. There should be no problems when creating new Series from a single series, but if we ever mix-and-match, then the two series must be ordered identically I believe

corebehrt/classes/outcomes.py
```diff
@@ -1,17 +1,15 @@
...
-import operator
 from datetime import datetime
-from typing import Dict, List, Tuple
+from typing import Dict, List, Optional, Set
 
 import numpy as np
 import pandas as pd
 
-from corebehrt.functional.filter import filter_events_by_abspos
+from corebehrt.common.constants import PID_COL, TIMESTAMP_COL
```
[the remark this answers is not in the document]
kirilklein: Opened #91

corebehrt/classes/outcomes.py
```diff
@@ -85,144 +83,70 @@ def match_concepts(
...
-            outcomes, start_followup_date, operator.ge
-        )
-        first_outcome = get_first_event_by_pid(filtered_outcomes)
-        return first_outcome
+        """Determine index dates based on mode."""
+        pids = set(patients_info[PID_COL].unique())
+
+        result = None
+        if index_date_mode == "absolute":
+            result = cls.create_timestamp_series(pids, index_date_params)
```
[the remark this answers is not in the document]
kirilklein: Fixed in 22f8375

corebehrt/classes/patient_filter.py
```diff
@@ -0,0 +1,129 @@
...
+
+    Args:
+        patients_info: DataFrame with PID column
+        initial_pids: Set of patient IDs to include
+        exposures: DataFrame with PID column
+        exposed_only: If True, keep only patients with exposures
+    """
+    if initial_pids:
+        patients_info = filter_df_by_pids(patients_info, initial_pids)
+    if exposed_only and exposures is not None:
```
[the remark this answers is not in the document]
kirilklein: Simplified in d01af89

corebehrt/classes/patient_filter.py
kirilklein: You're right. Maybe this even shouldn't be a flag at all, but just always remove

corebehrt/classes/prepare_data.py
kirilklein: I resolve this following our discussion

corebehrt/classes/patient_filter.py
kirilklein: 3ce3e6a

corebehrt/functional/outcomes.py
kirilklein: you're right, that is an outdated comment

corebehrt/classes/patient_filter.py
kirilklein: Removed it completely, since we never want to predict for patients who are already dead.

corebehrt/classes/outcomes.py
```diff
@@ -85,144 +83,70 @@ def match_concepts(
...
+        result.index.name = PID_COL  # Ensure index name is set to PID_COL
+        return result
 
     @staticmethod
     def draw_index_dates_for_unexposed(
         censoring_timestamps: pd.Series, data_pids: List[str]
     ) -> pd.Series:
-        """Draw censor dates for patients that are not in the censor_timestamps."""
+        """Draw censor dates for patients not in censor_timestamps."""
         np.random.seed(42)
```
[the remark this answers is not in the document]
mikkelfo: Open in #92

mikkelfo (approved):

PR #137, opened 2025-02-17
Azure logging

StephanLorenzen:
* Added utility functions for logging/plotting metrics in Azure (basically wrappers around mlflow logging utils).
* Added some logging in EHRtrainer and cross validation folds.
* Added mlflow + azure-mlflow to the `build_env` command in the azure sub-module.


StephanLorenzen:
2025-02-17-203443_657x409_scrot
2025-02-17-203508_673x551_scrot

How it looks with the division into folds. Currently each fold has only a single metric - I think, because there are basically no change to the accuracy/roc-auc during training because of the small amount of data for each fold when using the test data.

StephanLorenzen:
For pretraining, there is more data, and the graphs look nice:
2025-02-17-203810_1185x542_scrot

kirilklein (approved):
Looks good to me

Issue #146, opened 2025-02-18
Exclusion based on censoring


kirilklein:
Currently exclusion of prior outcomes is based on the index date. However, it would be more appropriate to based it on the censoring date instead. 
Otherwise for out of time evaluation we would still include dead patients for test and potentially validation.

PR #147, opened 2025-02-18
99 out of time

kirilklein:


kirilklein:
For OOT, different index dates are needed for train, val  and test.
We can introduce it in the cohort creation with minimal changes by providing optional params called shift_val and shift_test which will adjust the index date accordingly. This way the issue #146 is naturally solved.

kirilklein:
Problem: we do CV, so the censorig and would need to change for each fold.

kirilklein:
Instead of a validation script, we introduce a boolean parameter test which can be turned of for hyperparameter tuning and turned on for final model training.

kirilklein:
Test will not be created if predefined folds are provided but no test set.

kirilklein:
## PR Summary

- **Evaluation Flag:**  
  Added a new `evaluate` parameter in the configuration to control whether a separate test set should be used.

- **Flexible Test Set Handling:**  
  - When `evaluate` is true and predefined test PIDs are available, the code uses those to form the test set.  
  - If no predefined test PIDs exist and folds are not predefined, the code randomly splits the cohort into test and train/validation sets.  
  - If folds are predefined but no test set is provided, the pipeline runs cross-validation only.

- **Consistent Data Saving:**  
  Test PIDs are always saved to the processed data directory for consistency.

- **Improved Documentation:**  
  Added detailed comments within the finetune loop to clarify the rationale behind the evaluation and splitting logic.

These changes streamline the handling of test splits and cross-validation, ensuring that the evaluation strategy aligns with the provided configuration.

corebehrt/main/README.md
```diff
@@ -149,41 +149,88 @@ trainer_args:
...
+   To mimic a scenario where the model is trained with data up to the cutoff but then deployed later, we apply a shift (using `test_shift_hours`) exclusively to test patients. For example, with a 1 year test shift, training is performed using data up to January 1, 2020 (with outcomes censored relative to that date), while test patients are assigned a shifted index date of January 1, 2021. This simulates that the model is being applied to predict outcomes in a future time period.
+
+4. **Follow-up Window:**
+   The follow-up window is defined by the `n_hours_start_follow_up` and `n_hours_end_follow_up` parameters. For example, with a 1 year follow-up window, the model will predict outcomes in the period from January 1, 2020 to January 1, 2021 for train patients and from January 1, 2021 to January 1, 2022 for test patients.
+
+**Example Configuration:**
+
+- **Absolute Index Date:** January 1, 2020  
+- **test_shift_hours:** 365 * 24 (1 year)
+- During fine-tuning, we use the following configuration:
```
Montgomeryyyy, line 172: Which labels are the model then trained on? Also a year so that the input data is from 2019, and the labels are outcomes in 2020?

corebehrt/configs/finetune_oot.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+evaluate: true # evaluate best model on test sets
+
+model:
+  cls:
+    _target_: ehr2vec.model.heads.ClassifierGRU
+    bidirectional: true
+
+data:
+  test_split: 0.3 # only used if predefined_folds is false and test_pids is not provided
```
Montgomeryyyy, line 29: what if you want to use all patients in the test split? Do you then just use select_cohort to make the train/val split, and then use it again to make the test split?

Montgomeryyyy (changes requested):
Looks good, just a few comments about confusion of overall setup, which might be nice to include some comments/documentation about

corebehrt/main/README.md
kirilklein: added an additional explanation

corebehrt/configs/finetune_oot.yaml
kirilklein: It's now possible to provide all_pids as test_pids (just need to run select_cohort twice, once with test_split=1)

kirilklein: We will issue a warning that there is overlap

Montgomeryyyy (approved):
Ok

PR #148, opened 2025-02-19
fix: feature creation oom

kirilklein:


mikkelfo:
@kirilklein this can be closed for #151 right?

Issue #149, opened 2025-02-19
OOM during feature creation


kirilklein:
### Description
We run out of disk space when running create_data with a large enough dataset

### Solution
Save intermediate result to execute graph

kirilklein:
dask not used anymore

Issue #123, opened 2025-02-07
Bug when normalising values


Montgomeryyyy:
Currently a bug when normalising values. 
Solution could be so import Mikkels normalisation scheme, and only make binned normalisation available.

PR #151, opened 2025-02-24
109 meds loader

Montgomeryyyy:
Implemented a MEDS-loader for the data creation to fix OOM bugs (#149 ). 
Major changes include: 
* Data creation is now done using MEDS data. For this reason logic concerning splitting the data is moved to the MEDS library (MEDS_transform)
* Normalisation has been moved outside of create data, so this is now done in the ehr_preprocess pipeline, which is run before running MEDS_transform (addressing  #123 )
* All dask related function have now been replaced by pandas, and are performed shard-wise addressing the OOM issues in features. 
* Data preparation for training has been moved to a separate script, so that we do not need to run data preparation before each training run. 
* All names have been updated to align with the MEDS names
* All tests have been updated to the new names + MEDS data when necessary (#138 )

All the code has been tested on azure. Currently it takes 2.5-3 hours to create data . Pre-training has however slowed down as I can only use batch sizes of 64, I am not sure why this is so it would be nice if someone could take a look at it. 
Also right now we currently load in all the data before training. This seems to work for now, but we should probably come up with another solution like loading in with the dataloader, as I don't think this will work for long (maybe another issue)?


corebehrt/main/finetune_cv.py
```diff
@@ -32,7 +31,8 @@ def main_finetune(config_path):
     # Logger
     logger = logging.getLogger("finetune_cv")
 
-    data: PatientDataset = DatasetPreparer(cfg).prepare_finetune_data()
+    loaded_data = torch.load(join(cfg.paths.prepared_data, "patients.pt"))
```
kirilklein, line 34: Should be a constant

corebehrt/main/helper/create_data.py
```diff
@@ -11,52 +9,79 @@
...
-    concepts, patients_info = FormattedDataLoader(
-        cfg.paths.data,
-        cfg.loader.concept_types,
-        include_values=(getattr(cfg.loader, "include_values", [])),
-    ).load()
+    combined_patient_info = pd.DataFrame()
+    for split_name in ["train", "tuning", "held_out"]:
+        path_name = f"{cfg.paths.data}/{split_name}"
+        if not os.path.exists(path_name):
+            ValueError(f"Path {path_name} does not exist")
```
kirilklein, line 52: Should be `raise ValueError`

corebehrt/main/helper/create_data.py
```diff
@@ -11,52 +9,79 @@
...
+
+        df = tokenizer(df).reset_index()
+        os.makedirs(join(tokenized_path, f"features_{split}"), exist_ok=True)
+        df.to_parquet(
+            join(tokenized_path, f"features_{split}", f"{shard_n}.parquet"),
+            index=False,
+            schema=pa.schema(TOKENIZED_SCHEMA),
+        )
+        pids.extend(df["subject_id"].unique().tolist())
+    torch.save(set(pids), join(tokenized_path, f"pids_{split}.pt"))  # save pids as ints
```
kirilklein, line 42: As part of this PR or in a separate issue, making config a bit more specific helps. E.g. instead of passing config, pass features config and paths config. This makes it easier to track what is needed. Ideally, we pass the args directly eg. data_path, feat_path. If we have too many arguments, then it might be a sign that the function needs to be broken down further.

corebehrt/modules/features/excluder.py
```diff
@@ -11,6 +11,6 @@ def __init__(
         self.min_age = min_age
         self.max_age = max_age
 
-    def exclude_incorrect_events(self, df: dd.DataFrame) -> dd.DataFrame:
+    def exclude_incorrect_events(self, df: pd.DataFrame) -> pd.DataFrame:
```
kirilklein, line 14: Since this class is not doing anything except incorrect age exclusion, perhaps we should get rid of it and just use the function?

corebehrt/modules/features/features.py
```diff
@@ -70,27 +60,7 @@ def __call__(
...
-        concepts["ADMISSION_ID"] = concepts["ADMISSION_ID"].astype(str)
-        concepts["CONCEPT"] = concepts["CONCEPT"].astype(str)
-
-        patients_info = patients_info[["PID", "BIRTHDATE"]].compute()
-
-        def join_with_patients_info(concepts_partition: dd.DataFrame) -> dd.DataFrame:
-            """Join concepts with patients_info on PID."""
-            return concepts_partition.merge(patients_info, on="PID", how="left")
+        features = features.drop(columns=["admission_id", "time", "birthdate"])
+        features["subject_id"] = features["subject_id"].astype(str)
```
kirilklein, line 64: Why not keep it as int and save some space?

corebehrt/modules/features/loader.py
```diff
@@ -67,87 +63,80 @@ def read_file(file_path: str) -> pd.DataFrame:
...
+        filtered_schema = {
+            col: dtype for col, dtype in SCHEMA.items() if col in df.columns
+        }
+        df = df.astype(filtered_schema)
+        if TIMESTAMP_COL in df.columns:
+            df[TIMESTAMP_COL] = pd.to_datetime(df[TIMESTAMP_COL], errors="coerce")
+        return df
 
 
 class ParquetIterator:
```
kirilklein, line 142: Is this class needed?

corebehrt/modules/features/values.py
```diff
@@ -1,114 +1,38 @@
...
-        concepts = concepts.reset_index(drop=False)
-        concepts["index"] += 1
-        values = concepts.dropna(subset=["RESULT"])
-        if normalize_args is not None and callable(normalize_args["func"]):
-            values = normalize_args["func"](values, **normalize_args["kwargs"])
-
-        values["RESULT"] = (values["RESULT"] * 100).astype(int)
-        values["CONCEPT"] = "VAL_" + values["RESULT"].astype(str)
+    def bin_results(concepts: pd.DataFrame, num_bins=100) -> pd.DataFrame:
+        concepts["numeric_value"] = ValueCreator.bin(
```
kirilklein, line 12: This will likely trigger a warning in pandas. Better to make a copy and use .loc

setup.py
```diff
@@ -2,12 +2,13 @@
...
-      url="https://github.com/FGA-DIKU/EHR.git",
-      packages=['corebehrt'],
-     )
\ No newline at end of file
+setup(
+    name="corebehrt",
+    version="0.1",
+    description="",
+    long_description="",
+    authors="Kiril Klein, Mikkel Odgaard, Stephan Lorenzen",
```
kirilklein, line 10: +Maria Montgomery :D

corebehrt/constants/data.py
```diff
@@ -10,22 +10,26 @@
...
-PID_COL = "PID"
-TIMESTAMP_COL = "TIMESTAMP"
-BIRTHDATE_COL = "BIRTHDATE"
-DEATHDATE_COL = "DEATHDATE"
+PID_COL = "subject_id"
+TIMESTAMP_COL = "time"
+BIRTHDATE_COL = "birthdate"
+DEATHDATE_COL = "deathdate"
 ABSPOS_COL = "abspos"
+CONCEPT_COL = "code"
```
mikkelfo, line 18: Fine for now we call it "code", but it's ambiguous if we include values

corebehrt/functional/features/creators.py
```diff
@@ -152,36 +131,91 @@ def create_segments(concepts: dd.DataFrame) -> dd.DataFrame:
...
+
+    # Assign admission_id to all events between 'ADMISSION' and 'DISCHARGE' events
+    admission_mask = concepts["code"].str.contains("ADMISSION")
+    discharge_mask = concepts["code"].str.contains("DISCHARGE")
+    admission_indices = concepts[admission_mask].index.to_list()
+    discharge_indices = concepts[discharge_mask].index.to_list()
+
+    if admission_indices and discharge_indices:
+        start_idx = 0
+        while start_idx < len(admission_indices):
```
mikkelfo, line 166: Why is this not a for loop

corebehrt/functional/features/creators.py
```diff
@@ -152,36 +131,91 @@ def create_segments(concepts: dd.DataFrame) -> dd.DataFrame:
...
+            start = admission_indices[start_idx]
+            end = next((d for d in discharge_indices if d > start), None)
```
mikkelfo, line 168: A little freaky to do if our discharge_indices comes from a dataframe, cant we simply do `discharge_indices[discharge_indices > start].index[0]` (or something similar) and drop the conversion to list?

corebehrt/functional/features/creators.py
```diff
@@ -152,36 +131,91 @@ def create_segments(concepts: dd.DataFrame) -> dd.DataFrame:
...
+
+            if end:
+                concepts.loc[start:end, "admission_id"] = _get_adm_id()
+            start_idx += 1
+
+    # Assign admission_id to concepts outside admissions
+    # Concepts within 48 hours of each other are considered to be part of the same admission
```
mikkelfo, line 175: Is this new functionality or something that was in place?

corebehrt/functional/features/exclude.py
```diff
@@ -1,13 +1,16 @@
...
+) -> pd.DataFrame:
     """Exclude patients with incorrect ages (outside defined range)"""
     return df[(df["age"] >= min_age) & (df["age"] <= max_age)]
 
 
-def exclude_event_nans(df: dd.DataFrame) -> dd.DataFrame:
+def exclude_event_nans(df: pd.DataFrame, columns=None) -> pd.DataFrame:
     """Exclude events (row) with (any) NaNs"""
-    return df.dropna()
+    if columns:
```
mikkelfo, line 13: Dont need the if-statement here, since `subset=None` is default, so we can just do `df.dropna(subset=columns)`

corebehrt/functional/features/tokenize.py
```diff
@@ -22,9 +22,9 @@ def add_special_tokens_partition(
 
     if add_cls:
         # Find indices of the earliest event for each PID
-        cls_rows = df.groupby("PID").first()
+        cls_rows = df.groupby("subject_id").first()
         # Create [CLS] rows
-        cls_rows["concept"] = CLS_TOKEN
+        cls_rows["code"] = CLS_TOKEN
```
mikkelfo, line 27: Just leaving this comment here since I've seen it a lot of times, but perhaps we should've just added "code" as a constant and use this constant, rather than having to rename everything to "concept" and "subject_id"

corebehrt/functional/preparation/convert.py
```diff
@@ -4,6 +4,13 @@
 from tqdm import tqdm
 
 from corebehrt.modules.preparation.dataset import PatientData
+from corebehrt.constants.data import (
```
mikkelfo, line 7: This file uses `corebehrt.constants` but the others don't - all should use it

corebehrt/main/helper/create_data.py
```diff
@@ -11,52 +9,79 @@
...
-    )
-    torch.save(pids, join(tokenized_path, f"pids_{split}.pt"))
+    pids = []
+    for shard in os.listdir(join(features_path, split)):
+        shard_path = join(features_path, split, shard)
+        shard_n = shard.split(".")[0]
+        df = pd.read_parquet(shard_path).set_index("subject_id")
+
+        df = tokenizer(df).reset_index()
+        os.makedirs(join(tokenized_path, f"features_{split}"), exist_ok=True)
```
mikkelfo, line 33: If `tokenized_path` and `split` is pre-defined args, this should be outside the loop?

setup.py
mikkelfo, line 10: Should probably add the rest of you or just use a group name instead :)

mikkelfo (changes requested):
Some minor changes only, but there's two "requests":
1. We need to standardize using the constants, especially in a PR like this where we change "concept" to "code" and "subject_id"
2. We cannot merge this before we have figured out what is going on with the batch size

setup.py
kirilklein: We should switch to pyproject.toml anyway since the whole setup thing is outdated. But that's also for a separate PR

corebehrt/constants/data.py
kirilklein: It's for consistency with MEDS

.github/workflows/pipeline.yml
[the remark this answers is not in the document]
kirilklein: that's solved

mikkelfo:
> For the 2.: are we certain that the memory issue is introduced with this PR?
Yeah there is no issues on main with running larger batch sizes, so it's specific to this PR. Maria and I have been trying to do some debugging to figure out what is going on, but something is weird for sure.

corebehrt/configs/create_data.yaml
```diff
@@ -3,49 +3,26 @@ logging:
...
+    value_creator_kwargs:
+      num_bins: 100
 
 tokenizer:
   sep_tokens: true
   cls_token: true
-  #cutoffs:
-   # D: 3 # diagnosis
-   # M: 4 # medication
+  cutoffs:
```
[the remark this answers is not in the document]
Montgomeryyyy: I missed that. Do we already have this, or is this just another issue?

corebehrt/functional/features/creators.py
```diff
@@ -16,92 +19,69 @@ def create_abspos(concepts: dd.DataFrame, origin_point: datetime) -> dd.DataFram
...
+    adm_rows = concepts[
+        concepts["code"].str.contains("ADMISSION")
+        | concepts["code"].str.contains("DISCHARGE")
     ]
-
-    return background
-
-
-def assign_index_and_order(df: dd.DataFrame) -> dd.DataFrame:
+    concepts.loc[adm_rows.index, "code"] = "ADM_" + concepts.loc[adm_rows.index, "code"]
```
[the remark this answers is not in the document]
Montgomeryyyy: This was basically so that the cutoffs on  the diagnostic codes did not cutoff DISCHARGE. But if we are doing a mapping solution, then this also resolves this problem, and makes it so that we don't have to avoid cutoffs on "A"

corebehrt/main/helper/create_data.py
Montgomeryyyy: Yes, I think it makes sense to make this a separate issue. #158

corebehrt/modules/features/loader.py
Montgomeryyyy: I think it would make sense as we would probably load shards in multiple places with the MEDS format

corebehrt/configs/prepare_finetune.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+paths:
+## INPUTS
+  features: ./outputs/features
+  tokenized: ./outputs/tokenized
+  cohort: ./outputs/cohort # path to cohort directory
+  outcome: ./outputs/outcomes/TEST_OUTCOME.csv
+
+## OUTPUTS
+  prepared_data: ./outputs/finetuning/processed_data/ # Save model/outputs to this folder
```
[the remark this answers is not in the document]
Montgomeryyyy: #159

corebehrt/main/create_data.py
```diff
@@ -53,7 +50,7 @@ def main_data(config_path):
     # TODO: temporary fix/check until we split the script into two.
     # As cfg.paths.features is always set, its value cannot be used to decide
     # if features are present.
-    if os.path.exists(join(cfg.paths.features, "part.0.parquet")):
+    if os.path.exists(join(cfg.paths.features, "held_out", "0.parquet")):
```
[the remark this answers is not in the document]
Montgomeryyyy: I now made the held_out data optional.

corebehrt/functional/features/creators.py
Montgomeryyyy: As discussed today it has been like this for a while, since many concepts are not in an admission. I have now completely decoupled the admissions and the segments. I feel like that's what we agreed to, otherwise I'll just change it back.

Montgomeryyyy:
So I think I have now addressed most of the comments. 
I have tried standardising using the constants where I found issues, and also I can run it with larger batch sizes now (not sure why I couldn't before though)

kirilklein (approved):
Looks good to me!

corebehrt/modules/features/loader.py
mikkelfo: I mean since you made a `ShardLoader` will we not always use this rather than `ParquetIterator`?

mikkelfo (approved):
I'm fine with it as well - would probably just remove the cutoff thing (or comment it out) in the config, just so we don't change files that are not needed

PR #154, opened 2025-02-28
Async azure logging + tqdm postfix conflict

mikkelfo:
Synchronous logging had the GPU hanging, resulting in spiky GPU utilization

Setting tqddm's postfix as train loss ignores the `miniters` and `mininterval` and prints every tqdm update.

### DIFF
--- a/corebehrt/azure/log.py
+++ b/corebehrt/azure/log.py
@@ -6,6 +6,7 @@ try:
     # Try to import mlflow and set availability flag
     import mlflow
     import os
+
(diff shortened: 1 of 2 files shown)


kirilklein (approved):

mikkelfo:
Adding @StephanLorenzen for review so you can see how we'd do it now in azure

kirilklein (commented):
Looks fine to me, but why do we need to pass around the run object again? Can't we just access mlflow when necessary from inside the trainer or log function?

PR #157, opened 2025-03-03
Removed overwrite of restart_model if model_path had checkpoints

mikkelfo:


mikkelfo:
Not 100% sure what the expected behaviour is during fine-tuning, so I haven't touched this much. I only removed the functionality of "checkpoints", so `ModelManager` might have some redundancy in it in this PR

kirilklein (commented):
Looks good to me
As discussed, would be good to have a test for this

PR #162, opened 2025-03-04
Implemented autocasting (bfloat16) with GradScaler

mikkelfo:
Time save for pretrain train epoch (bs256, seq_len 512): 19 -> 12 minutes
Time save for pretrain val epoch (bs256, seq_len 512): 2.5 -> 2 minutes


kirilklein (approved):
looks good

Issue #168, opened 2025-03-05
First segment is set to 0, same as padding tokens


mikkelfo:
In the old code we do `factorize + 1`
https://github.com/mikkelfo/CORE-BEHRT/blob/7cf32d8dcec801a5d6bfdad055ae79aa36a401cf/ehr2vec/data/creators.py#L57

The new code has dropped the + 1 at some point, which makes the `segment=0` for padding tokens equal to the first segments
https://github.com/FGA-DIKU/EHR/blob/b51021e1a41cc56b420c7bee8e6d1e13eacac173/corebehrt/functional/features/normalize.py#L39

mikkelfo:
When tackling this issue, one should set `padding_idx` in `EhrEmbeddings` as well

mikkelfo:
We use the `padding_idx` in our `nn.Embedding`, but can't do it for `segment` due to this
https://github.com/FGA-DIKU/EHR/blob/1543aa668dd66eda65649abd1641583eb7f511be/corebehrt/modules/model/embeddings.py#L52-L55

padding_idx ([int](https://docs.python.org/3/library/functions.html#int), optional) – If specified, the entries at padding_idx do not contribute to the gradient; therefore, the embedding vector at padding_idx is not updated during training, i.e. it remains as a fixed “pad”. For a newly constructed Embedding, the embedding vector at padding_idx will default to all zeros, but can be updated to another value to be used as the padding vector.

kirilklein:
Also I guess now it would be a breaking change, since we then cannot run with existing tokenized features.

mikkelfo:
Yeah, I think that was the issue, that we'd have to re-run data processing, which I didn't want to do at the time - but I'm not 100% sure if that was the reason.

Issue #164, opened 2025-03-05
attention_mask is getting set as `torch.ones`


mikkelfo:
Currently, we pass in an attention_mask that includes padding https://github.com/FGA-DIKU/EHR/blob/cceb21611e1a001ea3278c4cab2a97fd582b767e/corebehrt/modules/model/model.py#L22 This makes the new ModernBERT unpadding approach unused, which should give a slowdown

kirilklein:
For the spedup to take effect, it needs FA2 for unpadding.

PR #169, opened 2025-03-05
Updated embeddings

### LINKED ISSUE #163: EhrEmbeddings does not require age/abspos as position_ids
`EhrEmbeddings` passes in `age` and `abspos` as `position_ids`, but now we explicitly pass in `input_embeds` to the model, so we don't need to wrap into `position_ids`

See [model](https://github.com/FGA-DIKU/EHR/blob/cceb21611e1a001ea3278c4cab2a97fd582b767e/corebehrt/modules/model/model.py) and [embeddings](https://github.com/FGA-DIKU/EHR/blob/main/corebehrt/modules/model/embeddings.py)

### LINKED ISSUE #166: Introduce `padding_idx` to `nn.Embedding`
Set the `padding_idx` in the `nn.Embedding` inside the `EhrEmbeddings` class which makes the padding_token a zero-vector. Might be redundant, but [ModernBERT does it](https://github.com/huggingface/transformers/blob/752ef3fd4e70869626ec70657a770a85c0ad9219/src/transformers/models/modernbert/modeling_modernbert.py#L202)

From https://pytorch.org/docs/stable/generated/torch.nn.Embedding.html: "If specified, the entries at `padding_idx` do not contribute to the gradient; therefore, the embedding vector at `padding_idx` is not updated during training, i.e. it remains as a fixed “pad”. For a newly constructed Embedding, the embedding vector at `padding_idx` will default to all zeros, but can be updated to another value to be used as the padding vector."

mikkelfo:
Closes #166 and closes #163, but is blocked by #168 (so `padding_idx` isn't set for `segment_embedding`)

### DIFF
--- a/corebehrt/modules/model/model.py
+++ b/corebehrt/modules/model/model.py
@@ -14,6 +14,7 @@ class BertEHREncoder(ModernBertModel):
             hidden_size=config.hidden_size,
             type_vocab_size=config.type_vocab_size,
             embedding_dropout=config.embedding_dropout,
+            pad_token_id=config.pad_token_id,
         )
 
     def forward(self, batch: dict):
--- a/corebehrt/modules/model/model.py
+++ b/corebehrt/modules/model/model.py
@@ -17,15 +17,13 @@ class BertEHREncoder(ModernBertModel):
         )
 
     def forward(self, batch: dict):
-        input_ids = batch["concept"]
-        token_type_ids = batch["segment"]
-        attention_mask = torch.ones_like(input_ids)
-        position_ids = {key: batch[key] for key in ["age", "abspos"]}
+        attention_mask = torch.ones_like(batch["concept"])
 
         inputs_embeds = self.embeddings(
-            input_ids=input_ids,
-            token_type_ids=token_type_ids,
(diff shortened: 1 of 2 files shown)


corebehrt/modules/model/model.py
```diff
@@ -14,18 +14,17 @@ def __init__(self, config):
...
             embedding_dropout=config.embedding_dropout,
+            pad_token_id=config.pad_token_id,
         )
 
     def forward(self, batch: dict):
-        input_ids = batch["concept"]
-        token_type_ids = batch["segment"]
-        attention_mask = torch.ones_like(input_ids)
-        position_ids = {key: batch[key] for key in ["age", "abspos"]}
+        attention_mask = torch.ones_like(batch["concept"])
```
kirilklein, line 21: Will you look at fixing the attention mask and unpadding as part of this PR?

kirilklein (approved):
Looks godd

corebehrt/modules/model/model.py
mikkelfo: No, I left it for another issue #164 and have a different branch for it - Don't want to overflow this one

Issue #170, opened 2025-03-05
Code crashes when selecting cv=1


Montgomeryyyy:
### Description
An error occurs if cv_folds is defined as 1 in select_cohort . 

### Logs/Traceback (if applicable)
  File "/mnt/azureml/cr/j/2d457d1b309845a3bd196738d922d9e4/exe/wd/corebehrt/functional/features/split.py", line 154, in create_folds
    kf = KFold(n_splits=num_folds, shuffle=True, random_state=seed)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/azureml-envs/azureml_d9ffa159a1e5e8b05a29397807dcfc71/lib/python3.12/site-packages/sklearn/model_selection/_split.py", line 514, in __init__
    super().__init__(n_splits=n_splits, shuffle=shuffle, random_state=random_state)
  File "/azureml-envs/azureml_d9ffa159a1e5e8b05a29397807dcfc71/lib/python3.12/site-packages/sklearn/model_selection/_split.py", line 355, in __init__
    raise ValueError(
ValueError: k-fold cross-validation requires at least one train/test split by setting n_splits=2 or more, got n_splits=1.

Montgomeryyyy:
fixed in #171

PR #171, opened 2025-03-05
Select cohort cv1

Montgomeryyyy:
Added a val_ratio in select_cohort to fix bug when calling select_cohort with one split (#170)


kirilklein (approved):
all good

PR #172, opened 2025-03-06
fix label input to sklearn confusion matrix

Montgomeryyyy:
sklearn's confusion matrix expects labels to negative labels before positives, so the switch made it seem like the there were an overweight of positives to negatives.


kirilklein (approved):
Looks good
Are we certain it's correct this time?

PR #175, opened 2025-03-10
fix: cutoff before truncation

kirilklein:


mikkelfo (approved):
Just add "and reduced default epochs" to the commit message so we know when it happened (just in case)

PR #178, opened 2025-03-10
Removed references to dask and replaced with pandas

mikkelfo:
Also removed some functions (related to dask) that are unused, would be good to double-check that they are indeed not needed anywhere (couldn't find references in the code)


kirilklein (approved):
Very good

PR #179, opened 2025-03-11
Implemented sparse prediction, renamed models, and simplified MLM head

mikkelfo:
It was not completely straight forward to use `ModernBERTForMaskedLM` and `ModernBERTForSentenceClassification` (as discussed with @kirilklein), as we need to pass in a `batch` to our `Encoder`. I think this is good enough for now

Pretraining time: 12 min -> 10 min


kirilklein (approved):
Looks good. Anything we should but are not using from the ModernBertForSequenceClassification?

mikkelfo:
> Looks good. Anything we should but are not using from the ModernBertForSequenceClassification?

@kirilklein Not really, since the main differences are the head and the forward pass (ModernBERT supports regression, single-label and multi-label classification). But since we have the `BiGRU` head I don't think there's anything appropriate to borrow

PR #180, opened 2025-03-11
Overview

kirilklein:
A high level overview is provided here.
A more detailed figure with all inputs/outputs for each script will be added to main/README.md in a separate PR.


mikkelfo (approved):
I'm not 100% sure if there's meant to be a difference between blue, black, orange, and red boxes, but I think the figure is a good starting point for an overview

PR #181, opened 2025-03-11
Get rid of origin_point and use unix_epoch instead

kirilklein:


kirilklein:
TODO: 
1. add scaling option to model config
2. test on azure

mikkelfo (approved):
A bit overloaded, as this isn't exclusively about origin_point, but there's also some refactoring - But I'm fine just doing multiple things, we just have to reflect it in the commit/merge message

kirilklein:
Summary:

### Major bug fixes

- Fix nan values in censoring 
- Fix out of memory during parallel processing by changing backed
- Apply binning only if numeric column is present
- Safe arg access in case features is not provided by cfg

### Use unix epoch 

- consistent conversion from datetime to hours since unix epoch
- Update tests
- Update test data
- allow to set time2vec scale and shifts through config

Issue #187, opened 2025-03-13
Remove Cohort as input to finetune


kirilklein:
Suggestion:
We can simplify the pipeline by removing cohort as input to fine-tune and instead save the splits with prepared_data since that always takes the cohort as input.

kirilklein:
@Montgomeryyyy what do you think?

PR #199, opened 2025-03-17
remove cohort as finetune input

Montgomeryyyy:
Removed cohort input from finetune_cv as this is already given in prepare_data  addressing #187 . 
The previous predefined_split is thus now set as default.


kirilklein (approved):
Looks all fine to me

PR #201, opened 2025-03-18
Added code filtering

Montgomeryyyy:
Added code filtering using regex to filter wanted/unwanted code. 
The filtering excludes all codes that match to the regex.


corebehrt/main/helper/create_data.py
kirilklein, line 1: The exclude_regex should be a state of FeatureCreator, validated in __init__ to avoid repeated checks in each call.

kirilklein: e.g. via 
```
def is_valid_regex(pattern):
    try:
        re.compile(pattern)
        return True
    except re.error:
        return False
```

corebehrt/functional/preparation/filter.py
kirilklein, line 1: can be simplified, by using the suggestion below

kirilklein (changes requested):
Looks good, just one minor suggestion

kirilklein (approved):
all good

PR #205, opened 2025-03-19
Layer freezing

kirilklein:


kirilklein:
Results on 3 months lung cancer prediction with base pretrained model
- all layers: 5:20/epoch. ROC AUC 0.72
- all layers but two top frozen: 3:20/epoch. ROC AUC: 0.71
- all layers but head: 3:05/epoch. ROC AUC 0.69

kirilklein:
# Add Layer Freezing with Adaptive Unfreezing

This PR adds two new training features to enhance transfer learning:

1. **Layer Freezing**: Selectively freeze bottom transformer layers during training
   - Add `n_layers_to_freeze` parameter to specify how many layers to freeze
   - Reduce overfitting and speed up training on smaller datasets

2. **Adaptive Unfreezing**: Automatically unfreeze all layers when performance plateaus
   - Enable with `unfreeze_on_plateau: true`
   - Configure sensitivity with `plateau_threshold` parameter
   - Reset early stopping patience with `reset_patience_after_unfreeze`

These features enable a two-stage training approach: initially focus on task-specific layers, then transition to full model fine-tuning when progress stalls, improving both training efficiency and model performance.

corebehrt/functional/trainer/log.py
```diff
@@ -0,0 +1,84 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def log_number_of_trainable_parameters(model):
```
mikkelfo, line 6: Can also use this for inspiration (from lightning)
https://lightning.ai/docs/pytorch/1.6.2/_modules/pytorch_lightning/utilities/model_summary.html#ModelSummary

corebehrt/configs/finetune.yaml
```diff
@@ -40,6 +40,14 @@ trainer_args:
   early_stopping: 20
   stopping_criterion: roc_auc
   
+  # Freezing options
```
mikkelfo, line 43: I think we should disable this by default?

mikkelfo (changes requested):
Looks fine. My only concern would be the `_check` functions in the `Trainer`, since we're setting a lot of `self.` attributes, so it makes it sort of hard to follow a stack trace. Idk a better way, but my only suggestion would be that `check` returns a boolean and then we specifically call a `unfreeze` function based on the boolean

corebehrt/configs/finetune.yaml
kirilklein: These configs are mostly for the pipeline test to explore as much functionality as possible. We should not use them as example configs. I think it would help if we provide separate example configs with good parameters

PR #225, opened 2025-03-31
Fix/azure metrics

### LINKED ISSUE #186: Metric Logging
Combined metrics are not logged properly:
Image
Also these metrics are missing from the fold metrics so it's impossible to see the current learning rate

### LINKED ISSUE #221: Is there a way to pass cfg to azure job description?
At the moment, in order to see the config of a job, we need to get to the output. It would be handy if the config would be passed to the description, so when opening the job one can directly see the config.
@StephanLorenzen

StephanLorenzen:
* Added corebehrt job config to description to azure job, so it is visible in the dashboard.
* Changed child-run logging to use prefix instead of actual child runs. (The old functionality can be re-enabled by changing the global var in `azure/util/log.py`). This setup allows for nice comparison graphs within an experiment.

2025-03-31-155439_1531x454_scrot
2025-03-31-155453_455x390_scrot
2025-03-31-155512_745x403_scrot

Closes #186
Closes #221

### DIFF
--- a/corebehrt/azure/util/azure.py
+++ b/corebehrt/azure/util/azure.py
@@ -21,7 +21,6 @@ def is_azure_available() -> bool:
 
     :return: True if available, otherwise False.
     """
-    global AZURE_AVAILABLE
     return AZURE_AVAILABLE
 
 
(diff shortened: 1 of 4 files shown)


kirilklein (approved):
Great!

Issue #230, opened 2025-04-05
exclude_regex does not exclude corresponding values if there are values


Montgomeryyyy:
### Description
When using the exclude regex ex with labs, the lab tokens are removed but not the corresponding values.

PR #231, opened 2025-04-06
bug fix in exclude regex

Montgomeryyyy:
When excluding certain rows that match the exclude_regex, the corresponding values were not excluded (#230 ). 
I have now moved the exclusion of codes before the expansion of numeric values to deal with this. Also added an option for ignoring held_out data when performing data_creation.


kirilklein (approved):
looks good

Issue #232, opened 2025-04-06
Average instead of concatenate embeddings


kirilklein:
The outputs of bigru are concatenated, resulting in double the hidden dimension for the output. I suggest averaging the forward and backward output instead for efficiency.
What do you think @mikkelfo @Montgomeryyyy ?

mikkelfo:
We double the hidden dimensionality to have `hidden_dim` for each direction. If we worry about efficiency, any aggregation (aside from sum) of the ones tested in corebehrt showed pretty promising results without the added complexity of BiGRU. I don't think it makes sense to change BiGRU in the name of efficiency unless it doesn't change performance. 

If there's no performance change in averaging the hidden dim, then we should 100% do it. Another alternative would be to keep the current implementation, but instead of using `hidden_dim` for forward and backward, we can just use `hidden_dim/2`

Montgomeryyyy:
I also agree that if there is no performance change, then it makes sense to do it

kirilklein:
Used hidden_dim/2

PR #240, opened 2025-04-17
Eval script

Montgomeryyyy:
This pull request introduces a new evaluation pipeline for fine-tuned models, along with various configuration updates and code improvements. The key changes include adding a new `evaluate_finetune` module, updating configurations for better directory structure and functionality, and enhancing the model's ability to handle embeddings during evaluation. Below is a breakdown of the most important changes grouped by theme.

### New Evaluation Pipeline
* Added a new `evaluate_finetune` module (`corebehrt/main/evaluate_finetune.py`) to handle the evaluation of fine-tuned models, including metrics computation and saving predictions.
* Introduced helper functions for evaluation, such as `inference_fold`, `compute_metrics`, and `get_sequence_length`, in `corebehrt/main/helper/evaluate_finetune.py`. These support model inference, metrics calculation, and sequence length extraction.
* Created a new Azure component for running the evaluation pipeline, including input/output specifications and integration with the `job` utility (`corebehrt/azure/components/evaluate_finetune.py`).

### Configuration Updates
* Added `evaluate_finetune.yaml` to configure the evaluation pipeline, specifying paths, batch size, metrics (ROC AUC and PR AUC), and other parameters.
* Introduced `select_cohort_held_out.yaml` and `prepare_held_out.yaml` configurations for handling held-out test data, including logging, paths, and data preparation settings. [[1]](diffhunk://#diff-647ac5d4eb02681b3df680741846a277908e2242229d8cba9f60be02fac0df06R1-R39) [[2]](diffhunk://#diff-f464d246bcb9964aa4cd5e6e28dc39de60d5aa08ea3f2626d471928916a86614R1-R27)
* Updated existing configurations (`prepare_finetune.yaml`, `select_cohort.yaml`, `select_cohort_absolute.yaml`, etc.) to improve directory structure and align with the new pipeline. [[1]](diffhunk://#diff-c7493b51e9bf5b41c24be0f6ed918345522214c8dcc990a103045a84d475dbc7L9-R9) [[2]](diffhunk://#diff-62fbeeb2eb8f10628b98e9ae12e66513993d296dacc4f5910a2b6113fb517926L15-R15) [[3]](diffhunk://#diff-58258f7fb41cb4f81df413a65574241524bfc599c649bfcb5c9349b2d957946aL14-R14)

### Code Enhancements
* Enhanced the `BiGRU` and classifier heads in `corebehrt/modules/model/heads.py` to support an optional `return_embedding` parameter for returning embeddings during evaluation. [[1]](diffhunk://#diff-130f856a406c984f1fa45c55219e0faccb626fa281bd7e9694ac3f290670ee04L11-R20) [[2]](diffhunk://#diff-130f856a406c984f1fa45c55219e0faccb626fa281bd7e9694ac3f290670ee04L28-R38)

### Metrics and Utility Functions
* Added new utility functions for ROC AUC and PR AUC metrics in `corebehrt/functional/utils/metrics.py`. These are used in the evaluation pipeline for performance measurement.

### Documentation and Miscellaneous
* Updated the `README.md` to include steps for running the evaluation pipeline, such as selecting cohorts and preparing held-out data.
* Removed the outdated `finetune_evaluate.yaml` configuration file.


Montgomeryyyy:
We discussed making it optional to also get the embeddings when running the evaluation script, but as it may differ which embeddings you could want I just added all of them (EHR, BERT, BIGRU), where the EHR and BERT are mean pooled. 
I'm not sure if this makes sense though. Maybe a better option would be to not pool at all, or just include the last embedding? Lmk what you think @kirilklein

corebehrt/configs/evaluate_finetune.yaml
```diff
@@ -0,0 +1,23 @@
+env: local
```
kirilklein, line 1: I don't think we need the env:local

corebehrt/functional/utils/metrics.py
kirilklein, line 1: why not use it directly from sklearn with the _target_ and initialize function?

corebehrt/modules/model/heads.py
```diff
@@ -47,5 +57,7 @@ def forward(
             :, 0, self.hidden_size :
         ]  # First output from the backward pass
         x = torch.cat((forward_output, backward_output), dim=-1)
+        if return_embedding:
```
kirilklein, line 60: usually one would return both, if return_embedding=True, but I guess we never use both simultaneously?

corebehrt/modules/model/model.py
```diff
@@ -38,6 +38,9 @@ def forward(self, batch: dict):
             abspos=batch["abspos"],
         )
 
+        if return_embeddings:
```
kirilklein, line 41: We often would want the encodings not the input_embeddings
for returning input_embeds only, we can as well just initialize the embeddings module on it's own and pass the input through it.

logs/evaluate.log
kirilklein, line 1: this was probably added by accident

docs/.DS_Store
kirilklein, line 1: should this be part of the commit?

kirilklein (changes requested):
Very good, but I would return these 2 embeddings:

- Output of the encoder
- Patient level embeddings of the head before the classification MLP
Both can be useful for different purposes. Embeddings before the encoder could also be nice to look at, but that should be part of a different script, where we just load in the Embeddings module of the pretrained model. Here, since it's in the context of finetune evaluation, the most relevant are the head patient-level embeddings in my opinion.
I don't think we need custom pooling. If this is required for some analysis, one can easily perform it on the saved embeddings.

You can also half the hidden_dim #232 as part of this PR

corebehrt/modules/model/heads.py
Montgomeryyyy: yeah right now we are not using both at the same time, but I could combine it, and then return both?

kirilklein (approved):
Nice!

PR #257, opened 2025-04-29
Feat/combined outcomes

kirilklein:
## Summary by CodeRabbit

- **New Features**
  - Introduced support for defining and detecting combination outcomes based on temporal relationships between multiple events (e.g., identifying when one event occurs within a specific time window before or after another event).
  - Added configuration options to specify the time window and timestamp source for combined outcomes.
  - Added a new combined outcome definition `TEST_COMBINED` for testing purposes.

- **Bug Fixes**
  - Improved handling of empty or missing data during outcome matching to ensure robust processing.

- **Tests**
  - Expanded and enhanced test coverage for outcome matching, including complex combination scenarios and edge cases.

### DIFF
--- a/corebehrt/configs/outcomes_test.yaml
+++ b/corebehrt/configs/outcomes_test.yaml
@@ -8,15 +8,27 @@ paths:
   features: ./outputs/features/
 
 outcomes:
-  TEST_OUTCOME: 
-    type: [code]
-    match: [['D10', 'D02']]
-    exclude: ['157141000119108']
+  TEST_OUTCOME:
+    type: [ code ]
+    match: [ [ 'D10', 'D02' ] ]
+    exclude: [ '157141000119108' ]
     match_how: startswith
     case_sensitive: true
   TEST_CENSOR:
-    type: [code]
-    match: [['M112']]
+    type: [ code ]
+    match: [ [ 'M112' ] ]
     match_how: startswith
     case_sensitive: false
-
+  TEST_COMBINED:
+    combinations:
+      primary:
+        type: [ code ]
+        match: [ [ 'D10' ] ] # Death code
+        match_how: startswith
+      secondary:
+        type: [ code ]
+        match: [ [ 'D02' ] ] # Heart infarct code
+        match_how: startswith
+      window_hours: 1000 # Look for heart infarct within 72 hours of death
+      direction: "before" # Heart infarct should come before death
+      timestamp_source: "primary" # Use death timestamp as the outcome timestamp
--- a/corebehrt/configs/outcomes_test.yaml
+++ b/corebehrt/configs/outcomes_test.yaml
@@ -23,13 +23,13 @@ outcomes:
     combinations:
       primary:
         type: [ code ]
-        match: [ [ 'D10' ] ]
+        match: [ [ 'M8' ] ]
         match_how: startswith
       secondary:
         type: [ code ]
-        match: [ [ 'D02', 'D01' ] ]
+        match: [ [ 'LAB', 'D5' ] ]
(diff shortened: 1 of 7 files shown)


corebehrt/modules/cohort_handling/outcomes.py
```diff
@@ -32,27 +43,57 @@ def __call__(
...
-            timestamps[ABSPOS_COL] = timestamps[ABSPOS_COL].astype(int)
-            timestamps[PID_COL] = timestamps[PID_COL].astype(int)
+                types = attrs["type"]
+                matches = attrs["match"]
+                if types == "patients_info":
+                    timestamps = self.match_patient_info(patients_info, matches)
+                else:
+                    timestamps = self.match_concepts(
+                        concepts_plus, types, matches, attrs
+                    )
```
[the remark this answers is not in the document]
kirilklein: Maybe we should remove the patients_info option alltogether since everything is an event now?

corebehrt/configs/outcomes_test.yaml
```diff
@@ -8,15 +8,28 @@ paths:
...
-    match: [['M112']]
+    type: [ code ]
+    match: [ [ 'M112' ] ]
     match_how: startswith
     case_sensitive: false
-
+  TEST_COMBINED:
+    combinations:
+      primary:
+        type: [ code ]
```
Montgomeryyyy, line 25: when using the combined outcomes only the concepts_plus data is used and not the patient_info. So I think either type: code should not be an option as it is the only one, or it should also support mixing with patient info, like medication + death, etc

corebehrt/modules/cohort_handling/outcomes.py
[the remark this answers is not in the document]
Montgomeryyyy: haha I like that code rabbit can do reactions. But yeah could make sense to do.

corebehrt/modules/cohort_handling/outcomes.py
```diff
@@ -64,18 +105,93 @@ def match_concepts(
...
         We then ensure all the types are true for a row by using bitwise_and.reduce. E.g. CONCEPT==COVID_TEST AND VALUE==POSITIVE
         """
+        # Handle empty DataFrame
+        if len(concepts_plus) == 0:
+            return create_empty_results_df()
+
+        # Make a copy to avoid modifying the original
+        filtered_concepts = concepts_plus.copy()
+
         if "exclude" in attrs:
```
Montgomeryyyy, line 115: what is the use-case for exclusion?

Montgomeryyyy (changes requested):
very cool, just minor things

corebehrt/configs/outcomes_test.yaml
kirilklein: The idea behind type was to be able to select numeric value as well. But that is not implemented currently. I didn't want to rewrite the whole thing for now. 
We can do something more general like I did [here](https://github.com/kirilklein/PHAIR_EHR/blob/8c338ed55cd401bc5ac57c41cd63fd7aed294bbd/corebehrt/modules/cohort_handling/advanced/extract.py) or just switch to ACES to get the predicates csv and then convert to the format we have.

kirilklein: For now I'll just drop patients_info. We only ever used it for death, but that is an event now as well.

corebehrt/modules/cohort_handling/outcomes.py
kirilklein: E.g. all cancers except breas cancer

PR #259, opened 2025-04-29
Feat/delayed censoring

kirilklein:
**Delayed censoring**
Can be used to censor e.g.  diagnoses differently since those are coded with a time delay. Or apply harsher censoring for certain code groups if there is leakage 

## Summary by CodeRabbit

- **New Features**
  - Introduced support for concept-specific delay periods when censoring patient data, allowing different concepts to have custom delay times before being included.
  - Added configuration options to specify delay patterns for concepts using regular expressions.

- **Tests**
  - Added comprehensive tests for delay-aware censoring and regex-based concept delay mapping to ensure correct functionality.

- **Documentation**
  - Improved module-level documentation for patient data filtering utilities.
  - Enhanced user documentation with explanations of concept-specific delay configuration and its impact on data censoring.


corebehrt/configs/prepare_finetune.yaml
```diff
@@ -3,24 +3,29 @@ logging:
...
-outcome: # we will convert outcomes to binary based on whether at least one outcome is in the follow up window
+outcome:
+  # we will convert outcomes to binary based on whether at least one outcome is in the follow up window
   n_hours_censoring: -10 # censor time after index date (negative means before)
   n_hours_start_follow_up: 1 # start follow up (considering outcomes) time after index date (negative means before)
   n_hours_end_follow_up: null # end follow up (considering outcomes) time after index date (negative means before)
+
+concept_pattern_hours_delay:
+  # Diagnosis concepts (matching strings starting with "D")
+  "^D": 72
```
Montgomeryyyy, line 31: would be nice with one comment explaining the meaning of the value, like "negative means before". 
Also, maybe we should write somewhere that diagnoses are delayed in SP by 72 hours if this is the general case? I'm assuming the delay is to ensure that diagnostic codes (which are typically delayed) are better included in the history instead of being uniformly removed in censoring. So if this is general for all codes, then it would make sense for everyone to use with it, where it should be addressed in the main documentation.

Montgomeryyyy (changes requested):
I think it looks good. Maybe just add a bit more documentation, as it seems like a utility that makes sense for everyone to use.

corebehrt/configs/prepare_finetune.yaml
kirilklein: Good point!
No, I don't think it's the case in general, depends a bit on the application on how to handle this, 
but I will add some documentation

kirilklein:
any idea why status is not reported?
Haven't changed anything on the tests

Montgomeryyyy:
> any idea why status is not reported? Haven't changed anything on the tests

no, everything looks normal. Maybe just try pushing an empty commit, and see if that triggers it

Montgomeryyyy (approved):

PR #291, opened 2025-06-07
289 censoring code

kirilklein:
## Summary by CodeRabbit

- **New Features**
  - Predict token is now appended at the censor date during patient data preparation, with relevant metadata included.

- **Refactor**
  - Centralized vocabulary and predict token loading for more consistent and efficient data preparation.
  - Updated function signatures to support enhanced censoring logic and token handling.

- **Bug Fixes**
  - Improved handling of warnings during regex filtering.

- **Tests**
  - Test cases updated for more realistic event times, ages, and to verify new censoring and predict token behaviors.

- **Style**
  - Minor whitespace and import order adjustments for improved code clarity.

### DIFF
--- a/corebehrt/functional/preparation/convert.py
+++ b/corebehrt/functional/preparation/convert.py
@@ -13,7 +13,7 @@ from corebehrt.constants.data import (
 )
 
 
-def dataframe_to_patient_list(df: pd.DataFrame, dob_token: int) -> List[PatientData]:
+def dataframe_to_patient_list(df: pd.DataFrame) -> List[PatientData]:
     """Convert a DataFrame containing patient data into a list of PatientData objects.
 
     Args:
@@ -44,8 +44,6 @@ def dataframe_to_patient_list(df: pd.DataFrame, dob_token: int) -> List[PatientD
         abspos_list = group[ABSPOS_COL].tolist()
         segments_list = group[SEGMENT_COL].tolist()
         ages_list = group[AGE_COL].tolist()
-        if dob_token not in concepts_list:
-            raise ValueError(f"DOB token {dob_token} not found in concepts list")
         # Create a PatientData instance
         patient = PatientData(
             pid=pid,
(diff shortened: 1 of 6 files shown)


corebehrt/functional/preparation/filter.py
```diff
@@ -109,6 +118,26 @@ def censor_patient_with_delays(
...
+    """
+    Appends a predict token to the patient's data.
+    """
+    if patient.dob is None:
+        raise ValueError(
+            f"Patient {patient.pid} is missing date of birth (dob) required for age calculation"
+        )
+    patient.concepts.append(predict_token_id)
+    patient.abspos.append(float(censor_date))
+    patient.segments.append(1)
```
Montgomeryyyy, line 138: why is segments set at 1?

Montgomeryyyy (changes requested):
looks good. Just one question about the segments, should it not just be the last segment?

mikkelfo (approved):
Looks good if we address Maria's comment (I agree segment=1 seems weird, unless `normalize_segment` catches it later?) and can maybe be simplified if we assume `DOB_abspos=abspos[0]`

corebehrt/functional/preparation/filter.py
kirilklein: would you suggest max(segments)+1 instead?

mikkelfo: It would be more semantically correct to have it be `max(segments) + 1` imo, just to keep the increasing nature of the segment

corebehrt/functional/preparation/convert.py
```diff
@@ -44,14 +44,16 @@ def dataframe_to_patient_list(df: pd.DataFrame) -> List[PatientData]:
...
+        if dob_token not in concepts_list:
+            raise ValueError(f"DOB token {dob_token} not found in concepts list")
         # Create a PatientData instance
         patient = PatientData(
             pid=pid,
             concepts=concepts_list,
             abspos=abspos_list,
             segments=segments_list,
             ages=ages_list,
+            dob=abspos_list[concepts_list.index(dob_token)],
```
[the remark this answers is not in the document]
kirilklein: good point, I'll drop the CLS from the start

PR #315, opened 2026-02-18
Implement Lightning

Sllambias:
Potential bugs:
- Sampler NEVER used.
- Scheduler warmup num epochs > total num epochs

Big changes:
- [x] Migrate to Lightning
- [x] Migrate to Hydra
- [x] Upgrade Python version to 3.12
- [x] Upgrade package versions
- [x] Remove requirements.txt (deprecated)
- [x] Remove CONSTANTS files. Either set these as defaults in the classes/functions or use the hydra configs to define constants so we only have defaults being set in Configs+Definitions and not Configs+Constants+Definitions.

Dataflow description:
Create Data:
MEDS -> TrainingData, ValData, TestData + FullPopulation

PT:
We take TrainingData + ValData and filter based on subjects in Population. By default FullPopulation but this can be changed for other cohorts.

Create Outcome:
MEDS -> OutcomeA_FullPopulation

FT: 
We take TrainingData + ValData and filter based on subjects in Outcome_for_Population. By default full population for outcome but can be changed for other cohorts. So you can e.g. have OutcomeA_FullPopulation and OutcomeA_OnlyFemales

`create_data` (important) changelog:
 - Aggregation and value creator is removed (should be done in preMEDS, dataset level)  -- only `exclude_regex` is kept
 - `patients_info` is not saved anymore
 - Codding mapping removed (can be made with a custom vocabulary)
 - CLS token removed (never used -- `[CLS]` token is kept in default vocab for prediction tokens)

### DIFF
--- a/bonsai/functional/outcomes.py
+++ b/bonsai/functional/outcomes.py
@@ -73,12 +73,9 @@ def binarize_outcomes(
     time_delta_datetime = outcomes["outcome_date"] - outcomes["index_date"]
     time_delta_hours = time_delta_datetime.dt.days * 24
 
-    if n_hours_end_include is None:
-        outcomes_in_prediction_window = n_hours_start_include <= time_delta_hours
-    else:
-        outcomes_in_prediction_window = (
-            n_hours_start_include <= time_delta_hours <= n_hours_end_include
-        )
+    outcomes_in_prediction_window = n_hours_start_include <= time_delta_hours
+    if n_hours_end_include is not None:
+        outcomes_in_prediction_window &= time_delta_hours <= n_hours_end_include
 
     outcomes["label"] = outcomes_in_prediction_window.astype(int)
     return (


bonsai/modules/datamodules/base.py
Sllambias: Remove this?

bonsai/configs/core/base_train.yaml
```diff
@@ -2,7 +2,7 @@ hydra:
   verbose: false
   output_subdir: hydra
   run:
-    dir: ${oc.env:BONSAI_MODELS}/${dataset}/${hydra:job.config_name}/${data.fold}
```
Sllambias: Are we sure about removing fold from dir name? How will finetuned models end up not overwriting eachother if doing cv?

bonsai/functional/features.py
```diff
@@ -0,0 +1,120 @@
...
+
+    features = df
+    features["age"] = compute_age(features, dob_info)
+    features = exclude_incorrect_event_ages(features)
+
+    features["abspos"] = compute_abspos(features["time"])
+
+    features = features.sort_values(["subject_id", "time"]).reset_index(
+        drop=True
+    )  # TODO: Needed in MEDS?
```
mikkelfo: Is `.reset_index(drop=True)` needed? I would think not

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -0,0 +1,128 @@
...
+        masking_ratio: float = 0.8,
+        masking_random_ratio: float = 0.1,
+        masking_ignore_special_tokens: bool = True,
+        cutoff_date: Optional[dict] = None,
+    ):
+        super().__init__(subjects, max_len, background_length, cutoff_date=cutoff_date)
+        self.vocabulary = vocabulary
+
+        self.masking_select_ratio = masking_select_ratio
+        self.masking_ratio = masking_ratio
```
mikkelfo: Might need to explain the difference between `masking_select_ratio` (% of tokens to corrupt) and `masking_ratio` (% of tokens to replace with `[MASK]`) or revise naming

bonsai/modules/networks/components/embeddings.py
```diff
@@ -40,53 +30,54 @@ def __init__(
...
         self.dropout = nn.Dropout(embedding_dropout)
 
         # Initialize embeddings
-        self.concept_embeddings = nn.Embedding(
+        self.code_embedding = nn.Embedding(
             vocab_size, hidden_size, padding_idx=pad_token_id
         )
-        self.segment_embeddings = nn.Embedding(type_vocab_size, hidden_size)
-        self.age_embeddings = Time2Vec(
+        self.segment_embedding = nn.Embedding(type_vocab_size, hidden_size)
```
mikkelfo: We should actually implement `padding_idx=pad_token_id` and then start `segments` from `1` instead of `0`

bonsai/modules/networks/components/embeddings.py
mikkelfo: Delete this

bonsai/modules/create_data.py
```diff
@@ -0,0 +1,86 @@
+import logging
+from pathlib import Path
+from typing import Optional
+import pandas as pd
+import pyarrow as pa
+from bonsai.functional.features import create_features
+from bonsai.functional.regex_utils import exclude_codes
+
+
+def drop_duplicates(df: pd.DataFrame) -> pd.DataFrame:
```
mikkelfo: Should we find another place for this in `bonsai.functional`?

bonsai/readme.md
mikkelfo: Add a `train.py` readme if we decide that `finetune.py` and `train.py` must be seperated

bonsai/run/finetune.py
```diff
@@ -0,0 +1,124 @@
...
+        scheduler_warmup_epochs=cfg.training.scheduler_warmup_epochs,
+        pos_weight=get_loss_weight(
+            cfg.training.loss_weight_function,
+            labels=train_labels,
+        ),
+    )
+
+    ckpt_callback = ModelCheckpoint(
+        dirpath=model_save_dir,
+        monitor="val/loss",
```
Katja-Jagd: Maybe add the monitor parameter into finetune.yaml unless we always use val/loss for fine-tuning ofc.

.github/workflows/pipeline.yml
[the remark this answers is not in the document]
Sllambias: Yea, we could save a lightweight ckpt and ft from that.

bonsai/functional/normalization.py
```diff
@@ -0,0 +1,20 @@
...
+
+def normalize_segments(segments: torch.Tensor) -> List[int]:
+    """Normalize a list of segment IDs to be zero-based and contiguous.
+
+    Takes a list of segment IDs that may have gaps or start from an arbitrary number,
+    and normalizes them to start from 0 and increment by 1 while preserving their relative order.
+
+    Example:
+        >>> normalize_segments([5, 5, 8, 10, 8])
+        [0, 0, 1, 2, 1]
```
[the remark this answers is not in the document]
Sllambias: So we might not have `[5, 5, 8, 10, 8]` but will still have e.g. `[5, 5, 8, 8, 10]`, right? The current code handles both cases correctly, so I dont see any reason to change anything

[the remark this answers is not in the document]
mikkelfo: Yeah, I only think the docstring is wrong -- the function works the same

PR #342, opened 2026-06-23
predict step

Montgomeryyyy:
## Summary by CodeRabbit

* **New Features**
  * Added full `predict`-stage support for running inference on a configurable held-out split after training.
  * You can optionally save per-sample prediction outputs and aggregated prediction metrics to CSV files.
  * Training and finetuning workflows now generate and persist predictions automatically when a prediction split is provided.  
* **Configuration**
  * Introduced a new `predict_split` path in `configs/finetune.yaml` for held-out evaluation.

### DIFF
--- a/bonsai/modules/datamodules/FinetuneDataModule.py
+++ b/bonsai/modules/datamodules/FinetuneDataModule.py
@@ -13,7 +13,7 @@ class FinetuneDataModule(L.LightningDataModule):
         self,
         path_train_data: str,
         path_val_data: str,
-        path_test_data: str,
+        path_predict_data: str,
         path_population: str,
         batch_size: int,
         num_workers: int,
@@ -21,13 +21,13 @@ class FinetuneDataModule(L.LightningDataModule):
         max_len: int,
         train_outcomes: Dict[int, dict],
         val_outcomes: Dict[int, dict],
-        test_outcomes: Dict[int, dict],
+        predict_outcomes: Dict[int, dict],
         train_sampler: Optional[WeightedRandomSampler] = None,
     ):
         super().__init__()
         self.path_train_data = path_train_data
         self.path_val_data = path_val_data
-        self.path_test_data = path_test_data
+        self.path_predict_data = path_predict_data
         self.population = pl.read_csv(path_population)
 
         self.batch_size = batch_size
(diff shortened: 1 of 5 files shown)


bonsai/modules/datamodules/FinetuneDataModule.py
```diff
@@ -77,6 +79,24 @@ def setup_fit(self):
             max_len=self.max_len,
         )
 
+    def setup_predict(self):
+        if self.path_test_data is None:
```
Sllambias: I think we should use the path_predict_data if we're using the predict hooks

bonsai/modules/lightningmodules/FinetuneModule.py
```diff
@@ -15,11 +19,15 @@ def __init__(
         learning_rate: float = 5e-4,
         optimizer_epsilon: float = 1e-6,
         scheduler_warmup_epochs: int = 0,
+        predictions_path: Optional[Path] = None,
+        test_metrics_path: Optional[Path] = None,
```
Sllambias: Should also be predict rather than test here

bonsai/modules/lightningmodules/FinetuneModule.py
```diff
@@ -80,6 +88,59 @@ def test_step(self, batch, batch_idx):
...
+            self.predictions.append(probs.squeeze(-1))
+        return {
+            "subject_id": batch["subject_id"],
+            "logit": logits.squeeze(-1),
+            "prob": probs.squeeze(-1),
+            "label": labels.squeeze(-1),
+        }
+
+    def on_predict_epoch_end(self) -> None:
+        if self.predictions_path is None and self.test_metrics_path is None:
```
Sllambias: I think we can boil this down to prediction_output_dir and just save both there every time and then remove all of the "if X is not None" so the full code runs every time without a lot of if's and but's

Sllambias (changes requested):
Looks very neat, comments are just about naming convention - "predict" rather than "test" if we're using predict hook, to not confuse arg names in case test hook is used in the future. And the other is to remove all the conditionals and just require 1 correct arg and run the entire code every time

Sllambias (approved):
🥇

PR #344, opened 2026-07-01
Revamped model architecture and implemented a flash_attn version

### LINKED ISSUE #343: `is_causal` does nothing
The `_update_attention_mask` is not being called in newer `transformer` version

mikkelfo:
Removed our dependency on `transformers` and `ModernBert` to enable more control. Our model is now either:

1. A native PyTorch version (see `network/components/blocks_nofa.py`)
2. A FlashAttention version (see `network/components/blocks.py`)

To avoid another dependency, I have made a `rope.py` that includes `RotaryPositionalEmbedding` from the `torchtune` library. 

NB: I have removed your `heads.py` file with the `BiGRU` implementation. Now our `BonsaiFinetune` directly targets the `[CLS]` tokens using a standard `nn.Linear` for pooling, such that a bidirectional and causal model uses the same head. 

The FlashAttention code is currently untested

## Summary by CodeRabbit

* **New Features**
  * Rebuilt the EHR model with a custom Transformer stack and rotary positional embeddings.
  * Added selectable attention (SDPA by default) with optional FlashAttention support via an extra dependency.
* **Bug Fixes**
  * Updated warmup/linear scheduling to use PyTorch `LinearLR` (and omits the scheduler when warmup is 0).
  * Improved pretraining loss reshaping by deriving the vocabulary dimension from logits.
* **Breaking Changes**
  * Updated model architecture/config schema (e.g., `max_seqlen`, `num_layers`, `dropout`, `attention_dropout`, `attn_type`, `predict_token_id`).
  * Finetuning now derives hyperparameters directly from the pretrain checkpoint.
* **Chores / Tests**
  * Refreshed example test configs and Hydra output paths; removed obsolete config-merging unit tests.


bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -28,14 +29,11 @@ def __init__(
         self.val_loss = nn.CrossEntropyLoss()
         self.val_metrics = self.configure_metrics("val")
 
-        hparams = model.config.to_dict()
```
Sllambias, line 31: This means that the hparams file we save will be quite empty. Not your "fault" as you remove a config that is deprecated, but we may want to think of other ways to populate the hparams file to make it meaningful again.

bonsai/modules/networks/components/blocks.py
```diff
@@ -0,0 +1,62 @@
+import torch.nn as nn
+from flash_attn.layers.rotary import RotaryEmbedding
+from flash_attn.modules.mha import FlashSelfAttention
+
+try:
+    from flash_attn.modules.mlp import FusedMLP as Mlp
```
Sllambias: What's the difference between these and why do we need a try except?

bonsai/modules/networks/components/blocks.py
```diff
@@ -0,0 +1,62 @@
...
+except Exception:
+    from flash_attn.modules.mlp import Mlp
+
+
+class FlashTransformerLayer(nn.Module):
```
Sllambias: is this different from a regular transformer layer? they look identical to me and the MLP also looks "normal" to me

bonsai/modules/networks/components/blocks_nofa.py
```diff
@@ -0,0 +1,95 @@
...
+            q,
+            k,
+            v,
+            attn_mask=attn_mask,
+            dropout_p=self.dropout if self.training else 0.0,
+            is_causal=self.causal and attn_mask is None,
+        )
+
+
+class TransformerLayer(nn.Module):
```
Sllambias: To me it looks like this layer could be shared across models, but I may be missing a detail

bonsai/modules/networks/components/embeddings.py
```diff
@@ -1,70 +1,41 @@
...
         vocab_size: int,
         hidden_size: int,
-        max_position_embeddings: int,
-        embedding_dropout: float,
-        pad_token_id: int = 0,
+        max_seqlen: int,
     ):
         super().__init__()
-        self.LayerNorm = nn.LayerNorm(hidden_size)
-        self.dropout = nn.Dropout(embedding_dropout)
```
Sllambias, line 33: not relevant anymore?

bonsai/modules/networks/bonsai_nets.py
```diff
@@ -1,118 +1,162 @@
...
-
-        Returns:
-            BaseModelOutput: output of ModernBertModel with last_hidden_state, etc.
-        """
-        inputs_embeds = self.embeddings(
-            input_ids=batch["code"],
-            segments=batch["segment"],
-            age=batch["age"],
-            abspos=batch["abspos"],
+        self.transformer = nn.ModuleDict(
```
Sllambias: I think these three should just be individual layers, they're not really interacting and also being called separately.

So we have self.transformer, self.dropout and self.layernorm

Sllambias (commented):
Codewise I think a big improvement with basically only positive changes. Major thing for me is to have this tested. Is that in the plan?

wrt naming I think it should be reversed so that the basic blocks are just called blocks and the FA blocks are called blocks_fa, rather than blocks_nofa, because negative naming is a difficult convention that scales poorly

bonsai/modules/networks/components/blocks.py
mikkelfo: `FusedMLP` requires additional packages (`csrc/fused_dense_lib`) to be compiled from their source and is not part of the base flash-attn. Maybe not worth the hassle; I'll just stick to the flash_attn `Mlp` instead

bonsai/modules/networks/components/blocks.py
mikkelfo: They're meant to be identical except that they use FlashAttention components. I can fuse it into one with `if else` swap from standard/flashattention components?

bonsai/modules/networks/components/embeddings.py
mikkelfo: `Dropout` is applied in `BonsaiBase.forward` instead, so that it lives inside the `self.transformer` stack. `LayerNorm` is redundant as modern transformers uses a pre-norm (´LayerNorm` is the first thing that happens in `TransformerLayer`)

mikkelfo:
@Sllambias @Montgomeryyyy The FlashAttention variant is untested; perhaps one of you can test it with the deep fetal data (also for a speed comparison)?

It can be pip installed through the optional dependecy `flash_attn` in the pyproject.toml and then toggling `attn_type="flash"` in the config

bonsai/modules/lightningmodules/FinetuneModule.py
```diff
@@ -150,10 +151,10 @@ def configure_optimizers(self):
         steps_per_epoch = (
             self.trainer.estimated_stepping_batches // self.trainer.max_epochs
         )
-        scheduler = get_linear_schedule_with_warmup(
+        scheduler = LinearLR(
             optimizer=optimizer,
-            num_warmup_steps=steps_per_epoch * self.scheduler_warmup_epochs,
-            num_training_steps=self.trainer.estimated_stepping_batches,
+            start_factor=1e-4,
+            total_iters=steps_per_epoch * self.scheduler_warmup_epochs,
```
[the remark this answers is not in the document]
Sllambias: this is a reasonable point for training without warmup

bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -87,10 +84,10 @@ def configure_optimizers(self):
...
             self.trainer.estimated_stepping_batches // self.trainer.max_epochs
         )
-        scheduler = get_linear_schedule_with_warmup(
+        scheduler = LinearLR(
             optimizer=optimizer,
-            num_warmup_steps=steps_per_epoch * self.scheduler_warmup_epochs,
-            num_training_steps=self.trainer.estimated_stepping_batches,
+            start_factor=1e-4,
+            total_iters=steps_per_epoch * self.scheduler_warmup_epochs,
         )
```
Sllambias, line 102: same here

bonsai/modules/networks/components/mha.py
```diff
@@ -0,0 +1,58 @@
...
+        self.dropout = dropout
+
+    def forward(self, q, k, v, attn_mask=None):
+        return F.scaled_dot_product_attention(
+            q,
+            k,
+            v,
+            attn_mask=attn_mask,
+            dropout_p=self.dropout if self.training else 0.0,
+            is_causal=self.causal and attn_mask is None,
```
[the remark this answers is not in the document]
Sllambias: maybe?

bonsai/modules/networks/components/mha_fa.py
```diff
@@ -0,0 +1,34 @@
...
+
+        qkv = self.Wqkv(x)
+        qkv = qkv.view(
+            bs, seqlen, 3, self.num_heads, self.head_dim
+        )  # (bs, seqlen, dim) -> (bs, seqlen, 3, nh, hd)
+
+        qkv = self.rotary_embedding(qkv, max_seqlen=self.max_seqlen)
+
+        y = self.self_attn(qkv, cu_seqlens=cu_seqlens, max_seqlen=self.max_seqlen)
+        y = y.reshape(bs, seqlen, -1)
```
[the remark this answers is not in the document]
Sllambias: ?

bonsai/modules/networks/bonsai_nets.py
```diff
@@ -1,118 +1,170 @@
...
-        self, attention_mask: torch.Tensor, output_attentions: bool
-    ) -> Tuple[torch.Tensor, torch.Tensor]:
-        global_attention_mask, sliding_window_mask = super()._update_attention_mask(
-            attention_mask, output_attentions
-        )
-        if self.is_causal:
-            global_attention_mask = make_attention_causal(global_attention_mask)
-            sliding_window_mask = make_attention_causal(sliding_window_mask)
+        # Weight tying (shares weights from code embedding to pretrain head)
+        self.embeddings.code_embedding.weight = self.pretrain_head.weight
```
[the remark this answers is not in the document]
Sllambias: ?

Sllambias (changes requested):
just a few coderabbit comments, ifthey're resolved then I think it's all good

Sllambias (approved):
:)

PR #346, opened 2026-07-25
Feat/vals

Montgomeryyyy:
- Inclusion of values using "FiLM" method from paper using the non-fused input from the ehr2meds pipeline
- Clean up in the example code and adding synthetic lab correlated code to test functionality of model for numeric signals

## Summary by CodeRabbit

* **New Features**
  * Added support for incorporating numeric clinical values alongside coded events.
  * Numeric values remain aligned through preprocessing, truncation, censoring, padding, and batching.
  * Added value-aware masked and autoregressive pretraining options.
  * Added configurable value regression loss with adjustable weighting.
  * Added example configurations for numeric-value pretraining, finetuning, and outcome validation.

* **Improvements**
  * Simplified configuration-driven model selection for training and finetuning.
  * Added checkpoint validation warnings for mismatched value-embedding settings.


bonsai/modules/networks/components/embeddings.py
```diff
@@ -19,14 +20,24 @@ def __init__(
...
         self,
         code: torch.LongTensor,
         age: torch.Tensor,
         abspos: torch.Tensor,
         segment: torch.LongTensor,
+        numeric_value: Optional[torch.Tensor] = None,
     ) -> torch.Tensor:
+
         embeddings = self.code_embedding(code)
+        if self.value_embedding_mode is not None and numeric_value is not None:
```
mikkelfo: Should probably make a double guard to prevent silent bugs where one exists and the other doesn't. Currently, a user can toggle one on (in the hopes of doing using values) but forget the other and have no way of knowing

bonsai/modules/networks/components/embeddings.py
```diff
@@ -19,14 +20,24 @@ def __init__(
...
+            embeddings = self.numeric_value_embedding(numeric_value, embeddings)
```
mikkelfo: Might be due to FiLM, so this is potentially incorrect, but I would've thought that it would be `embeddings += self.numeric_value_embedding(numeric_value)`

mikkelfo: Looking further, looks indeed to be FiLM-specific, so I guess this is of no concern. It's only a concern if we expect to support additional non-FiLM value embeddings; then this pattern breaks

bonsai/run/finetune.py
```diff
@@ -28,6 +28,15 @@
 load_dotenv()
 
 
+def resolve_value_embedding_mode(pretrain_hparams: dict, config_mode) -> str | None:
```
mikkelfo: Shouldn't this be automatically resolved? If any duplicate keys exist, then it should auto-fail if they are not equal? (I'm unsure whether the `.fit(ckpt_path)` uses `strict=True`)

mikkelfo (changes requested):
Overall, I think the code looks good from a functionality standpoint. However, there's one crucial thing that cements itself into two parts and that's all the `if` statement and optionals present.

1. @Sllambias would probably agree that this amount of toggling and optionals is a "code smell". I don't think we can fully avoid it, but neither do I think we should fully embrace it
2. Having `if` statements present in any model architecture (inside our model and nn.Modules), lead to graph breaks when using torch.compile. I.e. we're losing optimization when having this.

My suggestions would be: 
1. We get @Sllambias input and see if we can minimize the amount of optionals. This is mainly for all our functionals and data components. I can't really see many potential drops, but perhaps we can design something to avoid it being a general pattern
2. Make a seperate bonsai_net for anything to do with values. We cannot have `if` statements inside any `nn.Module`

bonsai/run/finetune.py
Montgomeryyyy: It didn't fail if I trained without value embeddings but added them in finetune, I just got a warning
Screenshot 2026-07-27 at 09 08 34

mikkelfo: I agree that we shouldn't pretrain using one version of value embeddings and finetune using another, but to me it seems okay (although perhaps not optimal) to pretrain _without_ value embeddings and finetune with them. Idk if you agree, but if you do, we could just check for `pretrain_mode is not None and pretrain_mode != finetune_mode`

Montgomeryyyy:
> Overall, I think the code looks good from a functionality standpoint. However, there's one crucial thing that cements itself into two parts and that's all the `if` statement and optionals present.
> 
> 1. @Sllambias would probably agree that this amount of toggling and optionals is a "code smell". I don't think we can fully avoid it, but neither do I think we should fully embrace it
> 2. Having `if` statements present in any model architecture (inside our model and nn.Modules), lead to graph breaks when using torch.compile. I.e. we're losing optimization when having this.
> 
> My suggestions would be:
> 
> 1. We get @Sllambias input and see if we can minimize the amount of optionals. This is mainly for all our functionals and data components. I can't really see many potential drops, but perhaps we can design something to avoid it being a general pattern
> 2. Make a seperate bonsai_net for anything to do with values. We cannot have `if` statements inside any `nn.Module`

Yeah I agree with the many if statements. Another alternative would be to add a nan numeric column, and then the code just runs as usual, since nan numerics are ignored. But then we would be carrying nan values throughout the model, which seems sub-optimal. I think that the separate model might make more sense

bonsai/run/finetune.py
Montgomeryyyy: Ah yeah I think that would be okay, I can change it to check the modes instead.

bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -69,16 +74,36 @@ def configure_metrics(self, prefix: str):
             ],
         )
 
+    def compute_combined_loss(self, batch, concept_loss_fn):
```
Sllambias: I think this should be a loss function specific to training with values that that is only instantiated with values so no conditionals are necessary

Sllambias (changes requested):
I agree with Mikkel and I see two options (but probably others are also possible):
1. Split into value specific funtions/classes that inherit the base and build upon. Downside of this is inheritance also becomes messy but if it's limited to 1 layer I think it can be workable.
2. Make values always included, e.g. as an empty list per subject. This would solve most of the conditionals and then only be required to split into separate classes at the model class, rather than having to make parallel classes from the dataset and forward. Would maybe also introduce some issues to have these, but is probably the simplest codewise if it is feasible.

Functionality looks good but I think it's better to do more proper review when the format is nailed down as things will probably change

bonsai/modules/lightningmodules/PretrainModule.py
Montgomeryyyy: I think (2) is definitely feasible as that's just what I ended up doing in the original codebase, but would we prefer this? Or something like (1) instead?

mikkelfo: (2) is probably the cleanest code-wise, but it's hacky, and (1) is probably the cleanest logic-wise. For this, I would try out (1) first and see if this resolves things. I don't mind having some if statements (e.g. `if "numeric_value" in subject`, then add it to the returned object).

So we could do either:
1. If logic is completely new, make a separate function/class for it
2. If this just extends functionality with another few lines of code, use inheritance or keep it as if statements. Here we could use inheritance for all classes and just if statements for functions.

Hard to tell if this also becomes messy in its own way, but I'm not the biggest fan of (2) unless alternatives don't work

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -94,8 +108,19 @@ def mask_patient_codes(
...
+        codes = codes.clone()
         codes[indices_mask] = self.vocabulary["[MASK]"]
 
+        if numeric_values is not None:
+            numeric_values = numeric_values.clone()
+            numeric_target = numeric_values.clone()
+            # Hide values only where the code is [MASK] and a real value existed.
+            numeric_value_masked = indices_mask & ~torch.isnan(numeric_values)
+            numeric_values[numeric_value_masked] = float("nan")
+            numeric_target[~numeric_value_masked] = float("nan")
```
Sllambias: Since we are replacing 'val' with 'nan' anyway can't we simplify to this :


```suggestion
        if numeric_values is not None:
            numeric_values[indices_mask] = float("nan")
            numeric_target[~indices_mask] = float("nan")
```

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -129,6 +154,16 @@ def __getitem__(self, index: int) -> dict:
         subject = super().__getitem__(index)
         subject["target"] = subject["code"][1:]
         subject["target"] = subject["target"].masked_fill(subject["target"] == 0, -100)
+
+        if "numeric_value" in subject:
+            values = subject["numeric_value"]
```
Sllambias: unnecessary I think

bonsai/modules/lightningmodules/PretrainModule.py
Sllambias: I think it's a bad idea to go forward with a _maybe it's there maybe it's not_ approach. That forces these lines in all functions, which is an eyesore in itself, and it also extends horribly:
```
if "numeric_value" in subject:
  subject["numeric_value"] = subject["numeric_value"].float()
```
I think either commit to it always being there, with the option of it being empty (solution 2), or commit to having specific functionality that should be invoked when values are there (solution 1)

Montgomeryyyy: So I split up the bonsai_nets to value and non-values, but there are still quite a lot of if-statements during data creation. Are we content with that, or would be also like to split this up? I'm not what the most clean way of doing this would be, as it is mostly just a few lines in each function.

bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -106,3 +114,58 @@ def configure_optimizers(self):
...
+            optimizer_epsilon=optimizer_epsilon,
+            scheduler_warmup_epochs=scheduler_warmup_epochs,
+        )
+        self.value_loss_fn = nn.MSELoss()
+
+    def add_hparams(self) -> dict:
+        return {"value_loss_weight": self.value_loss_weight}
+
+    def compute_loss(self, batch, concept_loss_fn):
+        logits, labels, val_logits, val_labels = self.model(batch)
```
Sllambias, line 150: This should be in the step functions I think

bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -106,3 +114,58 @@ def configure_optimizers(self):
...
+            learning_rate=learning_rate,
+            optimizer_epsilon=optimizer_epsilon,
+            scheduler_warmup_epochs=scheduler_warmup_epochs,
+        )
+        self.value_loss_fn = nn.MSELoss()
+
+    def add_hparams(self) -> dict:
+        return {"value_loss_weight": self.value_loss_weight}
+
+    def compute_loss(self, batch, concept_loss_fn):
```
Sllambias: The compute_loss should ideally be contained on the loss function entirely and only receive inputs directly needed by the loss function and this method would then be removed

configs/examples/example_outcome_val.yaml
Sllambias, line 1: What is the purpose of this example?

configs/pretrain.yaml
```diff
@@ -44,8 +45,9 @@ model:
   dropout: 0.1
   attention_dropout: 0.0
   bias: false
-  causal: false
+  causal: true
```
Sllambias, line 51: What does the causal change in both this and the finetune.yaml change?

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -70,14 +70,28 @@ def __init__(
```
Sllambias, line 70: I think the values warrant a dataset of their own. I don't think DataModule is necessary, we just need to make the dataset be configurable through the configs so the Values-configs can set the correct dataset class.

bonsai/run/finetune.py
```diff
@@ -28,6 +28,21 @@
...
```
Sllambias: I think this should just be set by the user and not inferred at all. If we want to be pedagogical we can check if something has changed between PT/FT in the load_checkpoint method of the nn.Module

bonsai/run/finetune.py
```diff
@@ -89,6 +107,12 @@ def main(cfg: DictConfig) -> None:
         attn_type=pretrain_cfg["attn_type"],
         predict_token_id=vocab["[CLS]"],
     )
+    if value_embedding_mode is not None:
```
Sllambias: I think the model class should be entirely configured in the config with the correct arguments, and not using kwargs + if-else patterns. We can use hydra's instantiate or similar to achieve this. Can also find an example if relevant

bonsai/run/pretrain.py
```diff
@@ -47,7 +50,7 @@ def main(cfg: DictConfig) -> None:
         max_len=cfg.training.max_len,
     )
 
-    model = BonsaiPretrain(
+    model_kwargs = dict(
```
Sllambias: same here as other file regarding kwargs + if-else

bonsai/run/train.py
```diff
@@ -68,7 +68,7 @@ def main(cfg: DictConfig) -> None:
         ),
     )
 
-    model = BonsaiFinetune(
+    model_kwargs = dict(
```
Sllambias: and here

Sllambias (changes requested):
Some where-should-this-go comments and then some putting stuff onto the user+configs and not into kwargs+if-else, but looks good with the network split I think

configs/examples/example_outcome_val.yaml
Montgomeryyyy: it's just the configuration needed to test the functionality of the values part of the code using the example data where the signal is in the continuous values. I have now renamed it to something more intuitive

configs/pretrain.yaml
Montgomeryyyy: I think this aligns better with the AR dataset (correct me if I'm wrong @mikkelfo )

bonsai/run/finetune.py
Montgomeryyyy: I changed this now, but there still is the pretrain model which is in an if-statement. Do we also want to move this to the config?

bonsai/functional/collate.py
mikkelfo, line 1: We could extend `dynamic_padding` to use an argument `extra_keys` and instantiate the function using a `functools.partial`. Idk if this might be more clean? @Sllambias

bonsai/functional/features.py
```diff
@@ -39,7 +41,10 @@ def create_features(df: pl.DataFrame) -> pl.DataFrame:
         )
     )
 
-    features = features.select("subject_id", "code", "age", "abspos", "segment")
+    select_cols = ["subject_id", "code", "age", "abspos", "segment"]
+    if numeric_column is not None:
+        select_cols.append(pl.col(numeric_column).alias("numeric_value"))
```
mikkelfo, line 46: Do we ever expect numeric_column to not be "numeric_value"? I can't decide if this is an unnecessary safeguard, a hidden unwanted interaction, or a good streamlining

bonsai/functional/subject_data.py
```diff
@@ -10,18 +10,20 @@ def prepare_subject_data(split_path: Path) -> List[Dict[str, torch.Tensor]]:
...
-            )
+            subject = {
+                "subject_id": subject_id[0],
+                "code": group["code"].to_torch(),
+                "abspos": group["abspos"].to_torch(),
+                "segment": group["segment"].to_torch(),
+                "age": group["age"].to_torch(),
+            }
+            if has_numeric:
+                subject["numeric_value"] = group["numeric_value"].to_torch().float()
```
mikkelfo: `.float` seems out of place when we don't do it for age and abspos and our `create_data` is already casting to float

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -81,22 +81,19 @@ def mask_patient_codes(
         target = codes.clone()
         probability_vector = torch.full(target.shape, self.masking_select_ratio)
 
-        # Ignore special tokens
```
mikkelfo, line 84: Idk if deleting the comments here is really necessary, might help to understand the code as it's a little obscure?

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -81,22 +81,19 @@ def mask_patient_codes(
...
-        # Get MLM mask
         selected_indices = torch.bernoulli(probability_vector).bool()
         target[~selected_indices] = -100
 
-        # Replace with [MASK]
         indices_mask = (
             torch.bernoulli(torch.full(target.shape, self.masking_mask_ratio)).bool()
             & selected_indices
         )
+        codes = codes.clone()
```
mikkelfo: Why do we clone here now?

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -113,6 +110,61 @@ def mask_patient_codes(
         return codes, target
 
 
+class ValueMLMPretrainDataset(MLMPretrainDataset):
+    def __getitem__(self, index: int) -> dict:
+        subject = super(MLMPretrainDataset, self).__getitem__(index)
```
mikkelfo, line 118: You can just call `super().__getitem__(index)` without the MLMPretrainDataset and self

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -113,6 +110,61 @@ def mask_patient_codes(
...
+        target[~selected_indices] = -100
+
+        indices_mask = (
+            torch.bernoulli(torch.full(target.shape, self.masking_mask_ratio)).bool()
+            & selected_indices
+        )
+        codes = codes.clone()
+        codes[indices_mask] = self.vocabulary["[MASK]"]
+
+        numeric_values = numeric_values.clone()
```
mikkelfo: Also dont think clone is needed here

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -113,6 +110,61 @@ def mask_patient_codes(
         return codes, target
 
 
+class ValueMLMPretrainDataset(MLMPretrainDataset):
```
mikkelfo, line 116: Overall, I think this dataset is wrongly put together. It gets the subject through inheritance, but then it goes through the exact same code again? To me, it almost looks like we're masking twice

mikkelfo: If the only difference between the two MLM and MLMwithValues is the 
```
        numeric_target = numeric_values.clone()
        # Hide input values at [MASK]; keep targets only at those positions.
        numeric_values[indices_mask] = float("nan")
        numeric_target[~indices_mask] = float("nan")
```
then we could just grab the `indices_mask` through the super call and delete most the code

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -129,6 +181,25 @@ def __getitem__(self, index: int) -> dict:
...
         subject["target"] = subject["target"].masked_fill(subject["target"] == 0, -100)
+
+        for key in ["code", "abspos", "segment", "age", "attention_mask"]:
+            subject[key] = subject[key][:-1]
+        return subject
+
+
+class ValueARPretrainDataset(ARPretrainDataset):
+    def __getitem__(self, index: int) -> dict:
+        subject = super(ARPretrainDataset, self).__getitem__(index)
```
mikkelfo: Can also swap to `super().__getitem__(index)` here

bonsai/modules/datasets/PretrainDataset.py
```diff
@@ -129,6 +181,25 @@ def __getitem__(self, index: int) -> dict:
         subject = super().__getitem__(index)
         subject["target"] = subject["code"][1:]
         subject["target"] = subject["target"].masked_fill(subject["target"] == 0, -100)
+
+        for key in ["code", "abspos", "segment", "age", "attention_mask"]:
+            subject[key] = subject[key][:-1]
+        return subject
+
+
+class ValueARPretrainDataset(ARPretrainDataset):
```
mikkelfo, line 163: Again, seems like the inheritance part goes wrong here and that we double apply the `[1:]` shift

bonsai/modules/lightningmodules/FinetuneModule.py
```diff
@@ -49,6 +50,18 @@ def __init__(
         )
         self.save_hyperparameters(hparams)
 
+    def on_load_checkpoint(self, checkpoint: dict) -> None:
+        """Warn on value_embedding_mode changes vs the pretrain ckpt."""
+        ckpt_mode = checkpoint.get("hyper_parameters", {}).get("value_embedding_mode")
+        ft_mode = self.model.hparams.get("value_embedding_mode")
+        if ckpt_mode != ft_mode:
```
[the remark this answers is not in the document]
mikkelfo: Agree with this. I think it's fine to only introduce `value_embedding_mode` during finetuning and not pretraining

bonsai/modules/networks/components/embeddings.py
```diff
@@ -26,15 +26,45 @@ def forward(
...
+class EhrValueEmbeddings(EhrEmbeddings):
+    def __init__(
+        self,
+        vocab_size: int,
+        hidden_size: int,
+        max_seqlen: int,
+        value_embedding_mode: str,
+    ):
+        super().__init__(vocab_size, hidden_size, max_seqlen)
+        self.value_embedding_mode = value_embedding_mode
```
mikkelfo: `EhrValueEmbeddings` doesn't seem to use `self.value_embedding_mode` (just passes it to `ContinuousEmbedding`, so we don't need to add it to `self`

bonsai/modules/networks/bonsai_nets.py
```diff
@@ -121,40 +170,81 @@ def __init__(
...
         self.pretrain_head = nn.Linear(hidden_size, vocab_size, bias=bias)
-
         # Weight tying (shares weights from code embedding to pretrain head)
         self.pretrain_head.weight = self.embeddings.code_embedding.weight
 
     def forward(self, batch: dict):
         last_hidden_state = super().forward(batch)
         labels = batch["target"]
 
-        # Predicts only on the non-masked tokens
```
mikkelfo, line 159: I don't see the need to delete all these comments that help explain the code

bonsai/modules/networks/bonsai_nets.py
```diff
@@ -87,21 +95,62 @@ def forward(self, batch):
...
+            abspos=batch["abspos"],
+            segment=batch["segment"],
+            numeric_value=batch["numeric_value"],
+        )
+
+
 class BonsaiPretrain(BonsaiBase):
     def __init__(
         self,
-        # Embedding / vocab
```
mikkelfo, line 124: Flavor related, but I think these comments slightly help understand the huge init statements (@Sllambias thoughts)?

bonsai/modules/networks/bonsai_nets.py
```diff
@@ -121,40 +170,81 @@ def __init__(
...
-        # Predicts only on the non-masked tokens
         mask = labels != -100
         last_hidden_state = last_hidden_state[mask]
         labels = labels[mask]
-
         logits = self.pretrain_head(last_hidden_state)
         return logits, labels
 
 
+class BonsaiValuePretrain(BonsaiValueBase):
```
mikkelfo: I feel like this class duplicates the `BonsaiPretrain` functionality. I dont know if double inheritance or something similar makes it more messy or cleaner, but right now we need to maintain the same code in two places (@Sllambias ?)

bonsai/modules/networks/bonsai_nets.py
```diff
@@ -174,11 +264,45 @@ def __init__(
...
-
-        # Extracts the hidden states corresponding to the predict token for each subject
         pred_tokens = batch["code"] == self.hparams["predict_token_id"]
         last_hidden_state = last_hidden_state[pred_tokens]
+        return self.finetune_head(last_hidden_state)
 
-        logits = self.finetune_head(last_hidden_state)
 
-        return logits
+class BonsaiValueFinetune(BonsaiValueBase):
```
mikkelfo: Same here as with the Pretrain version

bonsai/run/finetune.py
mikkelfo, line 1: Why is part of the kwargs deleted here?

bonsai/run/pretrain.py
```diff
@@ -47,18 +50,21 @@ def main(cfg: DictConfig) -> None:
...
-        max_seqlen=cfg.model.max_seqlen,
-        hidden_size=cfg.model.hidden_size,
-        num_layers=cfg.model.num_layers,
-        num_attention_heads=cfg.model.num_attention_heads,
-        bias=cfg.model.bias,
-        dropout=cfg.model.dropout,
-        attention_dropout=cfg.model.attention_dropout,
-        causal=cfg.model.causal,
-        attn_type=cfg.model.attn_type,
+    model = instantiate(cfg.model, vocab_size=len(data_module.vocabulary))
```
mikkelfo, line 48: The model instantiation differs from the `finetune.py` script and seems to have removed all the kwargs?

bonsai/run/train.py
mikkelfo, line 1: Again, why are all the kwargs deleted?

configs/pretrain.yaml
mikkelfo: The `ARPretrainDataset` is specified, so `causal` should be true for this

mikkelfo (changes requested):
I think there's a few things that we need to address/discuss:

1. Deletion of comments: A lot of comments has been deleted in the following code. I personally dislike this, as smaller comments can help with the code understanding and generally doesn't "cost" much. But again, code should be self-explanatory, so it's also a tradeoff. @Sllambias maybe grab a look and see what you think about this
2.  There has been changes to code logic (e.g. adding `.clone` statements) that shouldn't be necessary for an extension of the code
3. Model instantiation has changed and I believe the deletion of kwargs is wrong (if if not, less expressive which is also wrong)
4. The inheritance stuff likely has bugs and duplicates a lot of code

bonsai/functional/features.py
Montgomeryyyy: right now we load from "numeric_value_normalized", but I think it makes sense to make it adaptable for other columns if we want to load from binning or something later on.

bonsai/modules/datasets/PretrainDataset.py
Montgomeryyyy: I don't think we were masking twice, but I've now put them more together

bonsai/modules/lightningmodules/FinetuneModule.py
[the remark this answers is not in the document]
Montgomeryyyy: I think I prefer the message in general that something has changed, and then you can just ignore it if it is the intention (like the number of dropped duplicates, etc). But we can change it from a warning to just plain logging?

bonsai/run/finetune.py
Montgomeryyyy: I deleted these based on Sebastians comments (or my interpretation of them), that it would be cleaner to completely configure the model with the config instead of using kwargs @Sllambias

bonsai/run/pretrain.py
Montgomeryyyy: In the older code the instantiation of the finetune model was also based on the pretrain config, if that is what you mean?

bonsai/modules/datasets/PretrainDataset.py
Montgomeryyyy: I've added in the comments again for the "main" classes, but not for the duplicate code now

bonsai/run/pretrain.py
mikkelfo: More that we simply use `cfg.model` instead of passing the kwargs, but you answered that above

bonsai/modules/lightningmodules/FinetuneModule.py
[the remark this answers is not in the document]
Sllambias: I think it's fine "No Mode" is a mode that is different to "This Mode" and that's essentially what the warning is saying. Whether there should be a "no mode" is maybe another question but as long as that is a valid option I think this is fine

bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -15,6 +15,28 @@
     pass
 
 
+class ConceptValueLoss(nn.Module):
```
Sllambias: I think this is a nice change. Perhaps we can move it into a bonsai/modules/losses/ThisLoss.py file. And should it be called concept?

bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -106,3 +128,47 @@ def configure_optimizers(self):
             "frequency": 1,
         }
         return [optimizer], [scheduler_config]
+
+
+class ValuePretrainModule(PretrainModule):
```
Sllambias, line 127: It seems like this class can be entirely avoided if we:
1. Are content with just logging the loss value and
2. Add the loss functions and weights as configurable parts of the init of the PretrainModule.

That may perhaps be a more desirable compromise but I donno how important the logging of the different constituent loss values are

bonsai/modules/networks/components/embeddings.py
```diff
@@ -96,3 +125,37 @@ def forward(self, tau: torch.Tensor) -> torch.Tensor:
         periodic = self.f(linear_2 + self.phi)
 
         return torch.cat((linear_1, periodic), dim=-1)
+
+
+class ContinuousEmbedding(nn.Module):
```
Sllambias, line 130: I think this could easily just be hardcoded film layer as it doesn't seem like we're supporting anything else. And if we're intending to support other options then maybe separate embedding layers are a simpler approach

bonsai/modules/networks/bonsai_nets.py
Sllambias: I see what you mean. I think double inheritance is no good and also multiple layers of inheritance is also not nice.

Could we solve this by removing the BonsaiValueBase and inherting from the normal BonsaiBase and then changing the embed method and the init args here instead of in the BonsaiValueBase?

bonsai/modules/networks/bonsai_nets.py
Sllambias: Hm, I think I dont like these too much. I'd personally rather have more explicit arg names. E.g. attention_bias (like we already have attention_dropout) if a comment is needed already in your init. Alternatively a proper import string if explanation is due.

bonsai/modules/networks/bonsai_nets.py
Sllambias: And same suggested solution as with the Pretrain, if possible

bonsai/run/pretrain.py
Sllambias: I am happy to polish this as soon as I get back, I think the current solution is fine and almost achieves everything we would need to get it cleaned up completely :)

Sllambias (commented):
I think I am fine with the if-statements left in the functional parts. As long as we're only foreseeing this format for the values code and not expanding it to a bunch of other types of inputs etc., in which case the functions should be simplified and split up and referenced exactly by the classes/configs that uses them.

I also think the solution for the module instantiation is nice and very close to what I imagined with handling it directly from the configs with options for different arguments for different type of trainings, so that both if statements and kwargs are removed. And then I think the nicest solution for Maria is that I do the final polishing when I come back so that you can move and and until I come back from vacation I think the current solution is fine (if it works as intended) :)

bonsai/modules/lightningmodules/PretrainModule.py
Montgomeryyyy: Yes that sounds good, and I suppose it should be called CodeValueLoss or just combined loss or similar as think we are no longer using the "concepts" terminology anymore.

bonsai/modules/lightningmodules/PretrainModule.py
Montgomeryyyy: I think it would be nice to keep logging each part (at least since that's how I found issues in the training before), but making it configurable sounds like a good option.

bonsai/modules/networks/components/embeddings.py
Montgomeryyyy: Hmm my intention was that it should be easy to implement other options just by modifying the ContinousEmbedding, but I think whether or not we do that depends on the results we get on this vs just using discrete values on the real tasks using the real data. We could also just pin that and hardcode film, and then leave it up to the next person to change it should we want to support more options.

Montgomeryyyy:
> I think I am fine with the if-statements left in the functional parts. As long as we're only foreseeing this format for the values code and not expanding it to a bunch of other types of inputs etc., in which case the functions should be simplified and split up and referenced exactly by the classes/configs that uses them.
> 
> I also think the solution for the module instantiation is nice and very close to what I imagined with handling it directly from the configs with options for different arguments for different type of trainings, so that both if statements and kwargs are removed. And then I think the nicest solution for Maria is that I do the final polishing when I come back so that you can move and and until I come back from vacation I think the current solution is fine (if it works as intended) :)

Sounds good, I'll be happy if you want to take over for the final polishing:)

Sllambias:
> > I think I am fine with the if-statements left in the functional parts. As long as we're only foreseeing this format for the values code and not expanding it to a bunch of other types of inputs etc., in which case the functions should be simplified and split up and referenced exactly by the classes/configs that uses them.
> > I also think the solution for the module instantiation is nice and very close to what I imagined with handling it directly from the configs with options for different arguments for different type of trainings, so that both if statements and kwargs are removed. And then I think the nicest solution for Maria is that I do the final polishing when I come back so that you can move and and until I come back from vacation I think the current solution is fine (if it works as intended) :)
> 
> Sounds good, I'll be happy if you want to take over for the final polishing:)

Cool. Then I have nothing else to add and once my comments are resolved and/or addressed I think we're good to merge. Then I will pick up the polishing next week, but I don't think this needs to stay hanging until then.

mikkelfo:
I think I do not have any more comments of importance, so let's refer to @Sllambias for the last

bonsai/functional/collate.py
Montgomeryyyy: @Sllambias unsure what to do about this, and it may also be irrelevant with the unpadding PR?

bonsai/modules/lightningmodules/PretrainModule.py
Montgomeryyyy: I can take a look at this, but unsure if this is what you wanted to look at @Sllambias ?

Montgomeryyyy:
I've resolved most of the comments. There is just the pretrain module left, that I could combine to a single module possibly by instantiating the loss via the config or something similar? Unless this is what you wanted to take a look at @Sllambias

bonsai/functional/collate.py
mikkelfo: unpadding is implemented within the model's forward, so it shouldn't affect this.

I think we just keep it to the if-statements for now and then we can always extend it if we need (aka lets keep this one like it is)

Montgomeryyyy:
I think it is ready to be merged now, unless there is anything else you think needs to be adresses @mikkelfo and @Sllambias

Sllambias (approved):
lgtm :)

mikkelfo (approved):
Did a brief check, but lgtm

PR #347, opened 2026-07-30
Fix/345 unpadding

### LINKED ISSUE #345: Implement varlen (`cu_seqlens`) for the `attn_type=flash` toggle
Implement the unpadding/repadding approach for the `attn_type=flash` (is faster)

The functions `unpad_input` and `pad_input` can be found at https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/bert_padding.py

We simply need to call `unpad_input` BEFORE our forward pass and call `pad_input` AFTER

Katja-Jagd:
Closes issue #345

Implements unpadding for FlashAttention, packing sequences after the embedding layer and restoring the padded representation after the transformer layers.

The PR changes four files:
1. `bonsai_nets.py` unpads FlashAttention inputs after embeddings and repads the output after the transformer layers.
2. `layer.py` propagates `cu_seqlens` and `max_seqlen` through transformer layers.
3. `mha_fa.py` changes FlashAttention to operate directly on packed variable-length sequences.
4. `rope_fa.py` updates rotary embeddings to operate on packed variable-length QKV tensors

Performance testing was conducted on the PHAIR Light population on worker02, with runs limited to 500 training batches and 0 validation batches to keep the experiments short and consistent across the two implementations.

Performance was evaluated against the padded baseline using DeviceStatsMonitor, PyTorch Profiler, and iteration speed, but these measurements gave mixed results and did not show a clear performance advantage from unpadding. However, testing the maximum stable batch size showed a clear benefit: the unpadding implementation could handle larger batches before OOM, demonstrating reduced memory pressure and improved memory efficiency (See table below).

| Batch size  | Unpadding implemented | Unpadding not implemented |
|-----|-----------------------|---------------------------|
| 320 | —                     | ✅ PASSED                 |
| 352 | —                     | ✅ PASSED                 |
| 384 | —                     | ❌ FAILED OOM                |
| 512 | ✅ PASSED             | ❌ FAILED  OOM                |
| 640 | ✅ PASSED             | —                         |
| 768 | ❌ FAILED OOM            | —                         |

## Summary by CodeRabbit

* **Performance**
  * Improved attention processing for variable-length sequences when FlashAttention is available.
  * Reduced padding overhead by packing inputs before attention and restoring outputs afterward.
* **Compatibility**
  * Added support for standard scaled dot-product attention when FlashAttention is unavailable.
  * Added validation for unsupported attention configurations.
* **Reliability**
  * Improved handling of sequence metadata and attention masks across transformer layers.
  * Preserved dropout behavior across attention processing paths.

### DIFF
--- a/bonsai/modules/networks/bonsai_nets.py
+++ b/bonsai/modules/networks/bonsai_nets.py
@@ -7,6 +7,9 @@ from bonsai.modules.networks.components.layer import TransformerLayer
 
 _FLASH_ATTENTION_AVAILABLE = importlib.util.find_spec("flash_attn") is not None
 
+if _FLASH_ATTENTION_AVAILABLE:
+    from flash_attn.bert_padding import pad_input, unpad_input
+
 
 class BonsaiBase(nn.Module):
     def __init__(
@@ -73,19 +76,62 @@ class BonsaiBase(nn.Module):
             segment=batch["segment"],
         )
 
-        # SDPA requires a broadcasted attention mask
-        if self.hparams["attn_type"] == "sdpa" and not self.hparams["causal"]:
+        x = self.drop(x)
+
+        attn_type = self.hparams["attn_type"]
+        causal = self.hparams["causal"]
+
+        if attn_type == "flash":
+            # unpad_input expects shape (batch, seqlen) with
+            # 1/True for real tokens and 0/False for padding.
+            attention_mask = batch["attention_mask"].bool()
+
+            batch_size, padded_seqlen = x.shape[:2]
+
+            # x:
+            # (batch, padded_seqlen, hidden_size)
+            # -> (total_valid_tokens, hidden_size)
+            (
+                x,
+                indices,
+                cu_seqlens,
+                max_seqlen,
+                _,
+            ) = unpad_input(x, attention_mask)
+
+            for layer in self.layers:
+                x = layer(
+                    x,
+                    cu_seqlens=cu_seqlens,
+                    max_seqlen=max_seqlen,
+                )
+
+            x = self.layernorm(x)
+
+            # x:
+            # (total_valid_tokens, hidden_size)
+            # -> (batch, padded_seqlen, hidden_size)
+            x = pad_input(
+                x,
+                indices,
+                batch_size,
+                padded_seqlen,
+            )
+
+            return x
+
+        # Existing SDPA path.
+        if attn_type == "sdpa" and not causal:
             attn_mask = batch["attention_mask"][:, None, None, :]
         else:
             attn_mask = None
 
-        x = self.drop(x)
         for layer in self.layers:
             x = layer(
                 x,
                 attn_mask=attn_mask,
-                cu_seqlens=batch.get("cu_seqlens"),
             )
+
         x = self.layernorm(x)
 
         return x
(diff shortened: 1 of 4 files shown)


bonsai/modules/networks/bonsai_nets.py
```diff
@@ -73,21 +76,40 @@ def forward(self, batch):
...
         )
 
+        x = self.drop(x)
+
+        # FlashAttention requires unpadded input and cu_seqlens for variable-length sequences
+        if self.hparams["attn_type"] == "flash":
+            batch_size, padded_seqlen = x.shape[:2]
+            (x, indices, cu_seqlens, max_seqlen, _) = unpad_input(
+                x, batch["attention_mask"]
+            )  # x: (batch, padded_seqlen, hidden_size) -> (total_valid_tokens, hidden_size)
```
Sllambias: I would split these into separate methods from this point. So that the forward does the common steps and then calls:

```suggestion
        if self.hparams["attn_type"] == "flash":
            return forward_flash_attn(.....)
        elif self.hparams["attn_type"] == "sdpa" and not self.hparams["causal"]:
            return forward_sdpa(....)
```

With a bit fewer inline comments, e.g. type hints should be in the function/method definition and not where it is called. Maybe it would also be nice to expand the verbosity of the valueerror to explain what we expect and for which setups (i.e. sdpa is only valid with causal and flash can work with both. If flash is only intended to support causal or is causal by definition it should maybe also raise an error when flash=True and causal=False )

bonsai/modules/networks/components/mha_fa.py
```diff
@@ -23,19 +22,21 @@ def __init__(
...
         self.out_proj = nn.Linear(hidden_size, hidden_size, bias=bias)
 
-    def forward(self, x, cu_seqlens=None, **kwargs):
-        bs, seqlen, hidden_dim = x.shape
+    def forward(
+        self,
+        x,
+        cu_seqlens,
+        max_seqlen,
+        **kwargs,
```
Sllambias, line 30: Not your technical debt but I am not sure what kwargs we are supporting here. This seems like it may just be "error handling"

Sllambias (approved):
Very small comment from me, after which I think this is all good to go. But I'd appreciate a bit of inline-comment pruning and then the method split (without leading underscores in method definitions)

bonsai/modules/networks/bonsai_nets.py
mikkelfo: The only tradeoff is having duplicate code in both forward passes (the flow through the network is exactly the same)

bonsai/modules/networks/components/mha_fa.py
mikkelfo: It's because the `sdpa` version uses `attn_mask` and we use the same functions for both (in `layer.forward`). So we could also specify `attn_mask` for `flash` and `cu_seqlens/max_seqlen` for `sdpa`, but I feel like this causes more confusion than simply ignoring them through `kwargs`

bonsai/modules/networks/bonsai_nets.py
mikkelfo: Could be resolved through some inheritance where `flash` just extends the normal forward pass, but idk if this is better?

Sllambias: I think inheritance complicates this and the amount of duplicate code is very minimal so I think the simple solution is fine here

Sllambias: @mikkelfo you are happy with my solution, right?

mikkelfo (approved):

PR #348, opened 2026-08-12
upd

Sllambias:
The changes I wanted to make for Marias PR so they're merging with the feat/vals PR which would be merged into main right after this is merged (is the plan for me at least)

Streamlined loss functions to always do this:
1. Take inputs and targets as outputted by network
2. Internally handle the BONSAI shape and format to live up to torch format etc
3. Always output ONLY loss

@mikkelfo I changed how we do the loss imports also with flash attention, could you have a look? 

Also @Montgomeryyyy since I changed the loss functions to only output _the loss_ your logging is changed but I added the MSE metric and log that as part of the validation step (could also be in train step if necessary) which I think should give you essentially the same information as before but let me know if it's not as nice then I'll look at it again

Made a testing file which runs the whole values pipeline which I'll soon look to integrate into proper tests


bonsai/modules/lightningmodules/PretrainModule.py
```diff
@@ -70,15 +65,35 @@ def configure_metrics(self, prefix: str):
             ],
         )
 
+    def configure_losses(self, loss_fn, loss_params):
```
mikkelfo, line 68: Maybe this is easier to understand? We decide on `attn_type` then decide on loss

```
if self.model.hparams["attn_type"] == "sdpa":
    ce_fn = CE()
elif self.model.hparams["attn_type"] == "flash":
    from bonsai.modules.losses.CE_FA import CE_FA
    ce_fn = CE_FA()

if loss_fn == "CE":
    return ce_fn
elif loss_fn == "CodeValue":
    return CodeValueLoss(
        code_loss_fn=ce_fn,
        **loss_params,
    )
```

tests/run/values.py
```diff
@@ -0,0 +1,57 @@
+import subprocess
+import os
+
+os.environ["BONSAI_CONFIG_PATH"] = "../../configs"
```
mikkelfo, line 4: Can't we target something from the `env`?

tests/run/values.py
```diff
@@ -0,0 +1,57 @@
...
+        "bonsai/run/pretrain.py",
+        "--config-name",
+        "examples/example_pretrain_val",
+        "dataset=lab_correlated_MEDS_data_not_fused",
+    ],
+    check=True,
+    text=True,
+)
+
+# correct = subprocess.run(
```
mikkelfo: delete?

mikkelfo (approved):

Montgomeryyyy (approved):
looks good, and also runs fine on ngc on the fetal data.

bonsai/modules/lightningmodules/PretrainModule.py
Sllambias: Yea also had that in a draft but I decided against it because it becomes quite weird if we add just 1 more loss, imagine we have MSE loss and then we have to do this and the same for FA version of it and so forth, which leads to instantiating unnecessary losses:

````
if self.model.hparams["attn_type"] == "sdpa":
    ce_fn = CE()
    mse_fn = MSE()
```

tests/run/values.py
Sllambias: Yea, this was just a lazy implementation tbh because I want to do the workflow testing and that will require a different solution, so didn't wanna spend too much time on this when it's just a test and it runs now (but its very ugly yes)

PR #349, opened 2026-08-12
Added support for causal SDPA

mikkelfo:
Rather than forcing non-causal SDPA, we can just remove the `attn_mask` (SDPA doesn't allow `causal=True` and `attn_mask` to be set). 

Alternatively, we can have an if/else clause in the `forward_sdpa` that sets `attn_mask = None` instead of doing it in `mha`, but I don't believe the user (we) care for this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved causal attention handling for more reliable sequence processing.
  * Causal mode now correctly manages attention masks.
  * Invalid attention configurations continue to provide clear validation messages.

### DIFF
--- a/bonsai/modules/networks/bonsai_nets.py
+++ b/bonsai/modules/networks/bonsai_nets.py
@@ -32,6 +32,10 @@ class BonsaiBase(nn.Module):
             raise ImportError(
                 "flash_attn is not available. Please install flash-attn or use `attn_type='sdpa'` instead."
             )
+        if attn_type not in ["flash", "sdpa"]:
+            raise ValueError(
+                f"Invalid attn_type '{attn_type}'. Must be one of ['flash', 'sdpa']."
+            )
         super().__init__()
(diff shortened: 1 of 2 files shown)


bonsai/modules/networks/bonsai_nets.py
```diff
@@ -80,11 +80,11 @@ def forward(self, batch):
 
         if self.hparams["attn_type"] == "flash":
             return self.forward_flash(x, batch["attention_mask"])
-        elif self.hparams["attn_type"] == "sdpa" and not self.hparams["causal"]:
+        elif self.hparams["attn_type"] == "sdpa":
             return self.forward_sdpa(x, batch["attention_mask"])
```
Sllambias: could be handled by `attn_mask=batch["attention_mask"] if not self.hparams["causal"] else None`

Sllambias (approved):
Tiny suggested change to make it a little bit simpler and intuitive otherwise lgtm

